PR #3, opened 2016-09-07
Add logback-test.xml to reduce test output

poornachandra:


pom.xml
```diff
@@ -346,6 +352,18 @@
           </exclusion>
         </exclusions>
       </dependency>
+      <dependency>
+        <groupId>org.slf4j</groupId>
+        <artifactId>log4j-over-slf4j</artifactId>
+        <version>${slf4j.version}</version>
+        <scope>test</scope>
```
sagarkapare: why is this added in `test` scope only?

sagarkapare:
one comment else LGTM 👍

pom.xml
poornachandra: Since we don't know what kind of logging library will be used on the cluster, we only change it for test scope.

PR #4, opened 2016-09-07
TEPHRA-176 TEPHRA-177 : Adding maven modules for CDH-5.7 support, HBase-1.1 and HBase-1.2 modules.

sagarkapare:


tephra-examples/cdh-5.8/pom.xml
```diff
@@ -0,0 +1,96 @@
...
+
+  <build>
+    <sourceDirectory>../src/main/java</sourceDirectory>
+    <testSourceDirectory>../src/test/java</testSourceDirectory>
+  </build>
+
+  <dependencies>
+    <dependency>
+      <groupId>org.apache.tephra</groupId>
+      <artifactId>tephra-hbase-compat-1.2-cdh</artifactId>
```
poornachandra: Now that I think about it, it would be better to add a separate compat modules for CDH 5.7 and CDH 5.8. Since we now avoid code duplication for compat modules, it is just one extra jar we need to publish.

tephra-examples/pom.xml
```diff
@@ -52,7 +64,7 @@
     <dependency>
       <groupId>org.apache.hbase</groupId>
       <artifactId>hbase-common</artifactId>
-      <version>${hbase12.version}</version>
+      <version>${hbase11.version}</version>
```
poornachandra: HBase/Hadoop dependencies can be removed from the parent pom since the child poms will redefine them anyway, right?

tephra-core/src/test/resources/logback-test.xml
```diff
@@ -30,7 +30,7 @@
   <logger name="BlockStateChange" level="WARN" />
   <logger name="org.apache.zookeeper" level="WARN" />
   <logger name="org.mortbay.log" level="WARN" />
-  <logger name="org.apache.hadoop.hbase" level="INFO" />
+  <logger name="org.apache.hadoop.hbase" level="WARN" />
```
poornachandra: This change needs to be reverted.

tephra-hbase-compat-1.1-base/pom.xml
```diff
@@ -119,6 +121,12 @@
       <version>${hbase11.version}</version>
```
poornachandra: Same here, we should remove HBase/Hadoop dependencies from this parent pom too.

poornachandra:
@sagarkapare This changes will require update to documentation too for supported versions. That can be done in a separate PR.

tephra-examples/cdh-5.8/pom.xml
poornachandra: On second thoughts, this is not required. Since the HBase version is the same between CDH 5.7 and 5.8.

tephra-examples/pom.xml
sagarkapare: Fixed.

tephra-core/src/test/resources/logback-test.xml
sagarkapare: fixed.

tephra-hbase-compat-1.1-base/pom.xml
sagarkapare: Fixed.

sagarkapare:
Addressed review comments. @poornachandra please take a look again.

tephra-hbase-compat-1.1-base/pom.xml
```diff
@@ -0,0 +1,87 @@
...
+      <groupId>org.slf4j</groupId>
+      <artifactId>jcl-over-slf4j</artifactId>
+    </dependency>
+  </dependencies>
+
+  <profiles>
+    <profile>
+      <id>hbase1.2</id>
+      <properties>
+        <hbase11.version>1.2.1</hbase11.version>
```
poornachandra: This profile can be removed since we have a separate compat module for HBase 1.2 now

tephra-hbase-compat-1.1-base/tephra-hbase-compat-1.2/pom.xml
```diff
@@ -0,0 +1,118 @@
...
+    <version>0.9.0-incubating-SNAPSHOT</version>
+    <relativePath>../pom.xml</relativePath>
+  </parent>
+  <modelVersion>4.0.0</modelVersion>
+
+  <artifactId>tephra-hbase-compat-1.2</artifactId>
+  <name>Apache Tephra HBase 1.2 Compatibility</name>
+
+  <properties>
+    <hadoop.version>2.4.0</hadoop.version>
```
poornachandra: HBase 1.1 and 1.2 use 2.5.1 Hadoop version. It would be good to use the same version here, and for 1.1 compat module. We should also update the examples to use the same version.

poornachandra:
@sagarkapare Two more minor comments to address

tephra-hbase-compat-1.1-base/tephra-hbase-compat-1.2/pom.xml
sagarkapare: Updated the version for consistency.

tephra-hbase-compat-1.1-base/pom.xml
sagarkapare: Removed.

sagarkapare:
@poornachandra Addressed comments. Please take a look again. Thanks!

poornachandra:
LGTM. Please squash the commits.

PR #6, opened 2016-09-09
Prefer IPv4 stack while running tests to avoid HBase 1.0 startup falures in dual-stack nodes

poornachandra:


sagarkapare:
LGTM 👍

PR #7, opened 2016-09-09
Make TransactionManager Singleton for in-memory and local bindings.

anwar6953:
We don't have leader/follower logic happen in in-memory and local mode, so don't need TransactionManager to be singleton there.

This allows test cases that use the TransactionManager directly in the client to share the same instance.


poornachandra:
LGTM

PR #9, opened 2016-09-09
Wait for Transaction Service to announce itself to fix flaky tests

poornachandra:


tephra-core/src/test/java/org/apache/tephra/ThriftTransactionSystemTest.java
```diff
@@ -88,10 +89,14 @@ protected void configure() {
     txClient = injector.getInstance(TransactionSystemClient.class);
     try {
       LOG.info("Starting transaction service");
+      storage.startAndWait();
```
anwar6953: why start this here, but not in other test cases (TransactionAdminTest, for instance)?

tephra-examples/pom.xml
```diff
@@ -51,6 +55,53 @@
       <artifactId>tephra-core</artifactId>
       <version>${project.version}</version>
     </dependency>
+    <dependency>
```
anwar6953: why were all the pom changes necessary?

poornachandra: Adding HBase dependencies back to tephra-examples pom for IDE

tephra-core/src/test/java/org/apache/tephra/ThriftTransactionSystemTest.java
poornachandra: It was being stopped without being started, so for consistency I am starting it here. However, start on this object is a no-op.

anwar6953:
lgtm

PR #10, opened 2016-09-10
TEPHRA-179 Transaction service high availability changes

poornachandra:
Restructuring the Transaction Service classes to allow for HA restart while binding Transaction Manager and other classes as singletons


tephra-core/src/main/java/org/apache/tephra/distributed/TransactionService.java
```diff
@@ -42,28 +45,64 @@
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
+import javax.annotation.Nullable;
 
 /**
  *
  */
-public final class TransactionService extends InMemoryTransactionService {
+public class TransactionService extends AbstractService {
```
chtyim: Add a javadoc about this class to tell what does it do and when it should be used.

tephra-core/src/main/java/org/apache/tephra/distributed/TransactionService.java
```diff
@@ -42,28 +45,64 @@
...
+    // Retrieve the number of threads for the service
+    threads = conf.getInt(TxConstants.Service.CFG_DATA_TX_SERVER_THREADS,
+                          TxConstants.Service.DEFAULT_DATA_TX_SERVER_THREADS);
+    ioThreads = conf.getInt(TxConstants.Service.CFG_DATA_TX_SERVER_IO_THREADS,
+                            TxConstants.Service.DEFAULT_DATA_TX_SERVER_IO_THREADS);
+
+    maxReadBufferBytes = conf.getInt(TxConstants.Service.CFG_DATA_TX_THRIFT_MAX_READ_BUFFER,
+                                     TxConstants.Service.DEFAULT_DATA_TX_THRIFT_MAX_READ_BUFFER);
+
+    LOG.info("Configuring TransactionService" +
```
chtyim: Use the `{}` syntax instead of `+`.

tephra-core/src/main/java/org/apache/tephra/distributed/TransactionService.java
```diff
@@ -153,4 +193,31 @@ protected void internalStop() {
       }
     }
   }
+
+  private void undoRegister() {
```
chtyim: `unregister`?

tephra-core/src/main/java/org/apache/tephra/distributed/TransactionService.java
```diff
@@ -153,4 +193,31 @@ protected void internalStop() {
...
+    if (cancelDiscovery != null) {
+      cancelDiscovery.cancel();
+    }
+  }
+
+  private void doRegister() {
```
chtyim: `register`?

tephra-core/src/main/java/org/apache/tephra/runtime/DefaultTransactionManagerProvider.java
```diff
@@ -0,0 +1,71 @@
...
+  @Inject
+  public DefaultTransactionManagerProvider(Configuration conf, ZKClientService zkClientService) {
+    this.conf = conf;
+    this.zkClientService = zkClientService;
+  }
+
+  @Override
+  public TransactionManager get() {
+    // Create a new injector every time since Guice services cannot be restarted TEPHRA-179
+    Injector injector = Guice.createInjector(
```
chtyim: This is quite hacky in the way that usual provider doesn't create instance with a different injector. If all we need is a new instance, why not new it directly in here?

poornachandra: This will require more decoupling. We'll need to break the class hierarchy into HA Transaction Service, Transaction Service and Transaction Manager. For now I have added some get methods to help testing in PR https://github.com/apache/incubator-tephra/pull/11

poornachandra: Filed JIRA https://issues.apache.org/jira/browse/TEPHRA-182 for follow-up

PR #13, opened 2016-09-21
(Tephra-184) Remove unused constants

zevada:
These constants seem to have never been used. I've searched code going as far back as the initial commit of Tephra the apache repository.


poornachandra:
LGTM, thanks for the contribution!

PR #14, opened 2016-09-22
(TEPHRA-185) Add a way to pass a custom timeout to TransactionContext

anew:
Adds a new start(int timeout) to TransactionContext.

### DIFF
--- a/tephra-core/src/main/java/org/apache/tephra/TransactionContext.java
+++ b/tephra-core/src/main/java/org/apache/tephra/TransactionContext.java
@@ -89,6 +89,36 @@ public class TransactionContext {
    */
   public void start() throws TransactionFailureException {
     currentTx = txClient.startShort();
+    startAllTxAwares();
+  }
+
+  /**
+   * Starts a new transaction.  Calling this will initiate a new transaction using the {@link TransactionSystemClient},
+   * and pass the returned transaction to {@link TransactionAware#startTx(Transaction)} for each registered
+   * TransactionAware.  If an exception is encountered, the transaction will be aborted and a
+   * {@code TransactionFailureException} wrapping the root cause will be thrown.
+   *
+   * @param timeout the transaction timeout for the transaction
+   *
+   * @throws TransactionFailureException if an exception occurs starting the transaction with any registered
+   *     TransactionAware
+   */
+  public void start(int timeout) throws TransactionFailureException {
+    currentTx = txClient.startShort(timeout);
+    startAllTxAwares();
+  }
+
+  /**
+   * This is a helper for {@link #start()} and {@link #start(int)}.
+   *
+   * Passes the current transaction to {@link TransactionAware#startTx(Transaction)} for each registered
+   * TransactionAware. If an exception is encountered, the transaction will be aborted and a
+   * {@code TransactionFailureException} wrapping the root cause will be thrown.
+   *
+   * @throws TransactionFailureException if an exception occurs starting the transaction with any registered
+   *     TransactionAware
+   */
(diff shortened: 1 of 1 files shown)


tephra-core/src/main/java/org/apache/tephra/TransactionContext.java
```diff
@@ -88,7 +88,27 @@ public boolean removeTransactionAware(TransactionAware txAware) {
    *     TransactionAware
    */
   public void start() throws TransactionFailureException {
-    currentTx = txClient.startShort();
+    txClient.startShort();
```
poornachandra: `currentTx` still needs to be assigned the return value of `txClient.startShort()`, right?

tephra-core/src/main/java/org/apache/tephra/TransactionContext.java
```diff
@@ -88,7 +88,27 @@ public boolean removeTransactionAware(TransactionAware txAware) {
...
+   *
+   * @throws TransactionFailureException if an exception occurs starting the transaction with any registered
+   *     TransactionAware
+   */
+  public void start(int timeout) throws TransactionFailureException {
+    txClient.startShort(timeout);
+    staretAllTxAwares();
+  }
+
+  private void staretAllTxAwares() {
```
poornachandra: Typo in `staret`

tephra-core/src/main/java/org/apache/tephra/TransactionContext.java
```diff
@@ -88,7 +88,27 @@ public boolean removeTransactionAware(TransactionAware txAware) {
...
+   * TransactionAware.  If an exception is encountered, the transaction will be aborted and a
+   * {@code TransactionFailureException} wrapping the root cause will be thrown.
+   *
+   * @param timeout the transaction timeout for the transaction
+   *
+   * @throws TransactionFailureException if an exception occurs starting the transaction with any registered
+   *     TransactionAware
+   */
+  public void start(int timeout) throws TransactionFailureException {
+    txClient.startShort(timeout);
```
poornachandra: Same here about assignment

tephra-core/src/main/java/org/apache/tephra/TransactionContext.java
```diff
@@ -97,6 +127,7 @@ public void start() throws TransactionFailureException {
                                        txAware.getTransactionAwareName(), currentTx.getTransactionId());
         LOG.warn(message, e);
         txClient.abort(currentTx);
+        currentTx = null;
```
anew: this is not strictly part of this new feature. Just fixing this along the way. Normally, after we abort the tx, we set currentTx = null. I thought it would be best to do that here, too

tephra-core/src/main/java/org/apache/tephra/TransactionContext.java
anew: oh yes. fixed

tephra-core/src/main/java/org/apache/tephra/TransactionContext.java
anew: ditto

tephra-core/src/main/java/org/apache/tephra/TransactionContext.java
anew: oops.

anew:
@poornachandra please take another look.

tephra-core/src/main/java/org/apache/tephra/TransactionContext.java
poornachandra: 👍

PR #17, opened 2016-10-06
(TEPHRA-188) Allow to configure a limit for the transaction timeout

anew:
the title says it all.

### DIFF
--- a/tephra-core/src/main/java/org/apache/tephra/TransactionManager.java
+++ b/tephra-core/src/main/java/org/apache/tephra/TransactionManager.java
@@ -142,6 +142,7 @@ public class TransactionManager extends AbstractService {
   private final int cleanupInterval;
   private final int defaultTimeout;
   private final int defaultLongTimeout;
+  private final int maxTimeout;
   private DaemonThreadExecutor cleanupThread = null;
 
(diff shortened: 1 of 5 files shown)


tephra-core/src/main/java/org/apache/tephra/TransactionManager.java
```diff
@@ -722,7 +725,10 @@ public Transaction startShort() {
    * @param timeoutInSeconds the time out period in seconds.
    */
   public Transaction startShort(int timeoutInSeconds) {
-    Preconditions.checkArgument(timeoutInSeconds > 0, "timeout must be positive but is %s", timeoutInSeconds);
+    Preconditions.checkArgument(timeoutInSeconds > 0,
+                                "timeout must be positive but is %s", timeoutInSeconds);
+    Preconditions.checkArgument(timeoutInSeconds <= maxTimeout,
+                                "timeout must not exceed %s but is %s", maxTimeout, timeoutInSeconds);
```
poornachandra: It would be good to add the units here

anew:
@poornachandra I made the change. Please take another look.

poornachandra:
LGTM

PR #18, opened 2016-10-20
[TEPHRA-194] Make startShort() throw IllegalArgumentException …

anew:
… consistently for invalid timeout values.

Previously, it was throwing IllegalArgumentException in-memory and Exception for thrift mode. 

This breaks down into:
- adding a new exception to the thrift stubs to be able to communicate back any exception
- adding a new method startShortWithTimeout that throws that exception
- modifying the TransactionServiceThriftClient to convert that exception back into an IllegalArgumentException
- modifying the TransactionSystemTest to expect IllegalArgumentException now
- modifying the ThriftTransactionServerTest to use a retry strategy that counts retries and validate that the number of retries is 0 if startShort() throws IllegalArgumentException
- modifying TransactionServiceClient to accept a class name as the retry strategy provider (needed for above test case)
- some unrelated but reasonable style fixes

### DIFF
--- a/tephra-core/src/main/java/org/apache/tephra/distributed/TransactionServiceThriftClient.java
+++ b/tephra-core/src/main/java/org/apache/tephra/distributed/TransactionServiceThriftClient.java
@@ -25,6 +25,7 @@ import org.apache.tephra.InvalidTruncateTimeException;
 import org.apache.tephra.Transaction;
 import org.apache.tephra.TransactionCouldNotTakeSnapshotException;
 import org.apache.tephra.TransactionNotInProgressException;
+import org.apache.tephra.distributed.thrift.TGenericException;
 import org.apache.tephra.distributed.thrift.TInvalidTruncateTimeException;
 import org.apache.tephra.distributed.thrift.TTransactionCouldNotTakeSnapshotException;
 import org.apache.tephra.distributed.thrift.TTransactionNotInProgressException;
@@ -33,6 +34,8 @@ import org.apache.thrift.TException;
 import org.apache.thrift.protocol.TBinaryProtocol;
 import org.apache.thrift.protocol.TProtocol;
 import org.apache.thrift.transport.TTransport;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import java.io.ByteArrayInputStream;
 import java.io.InputStream;
@@ -54,6 +57,8 @@ public class TransactionServiceThriftClient {
       return ByteBuffer.wrap(input);
     }
   };
+
+  private static final Logger LOG = LoggerFactory.getLogger(TransactionServiceThriftClient.class);
 
   /**
    * The thrift transport layer. We need this when we close the connection.
@@ -112,7 +117,15 @@ public class TransactionServiceThriftClient {
 
   public Transaction startShort(int timeout) throws TException {
     try {
-      return TransactionConverterUtils.unwrap(client.startShortTimeout(timeout));
+      return TransactionConverterUtils.unwrap(client.startShortWithTimeout(timeout));
+    } catch (TGenericException e) {
+      // currently, we only expect IllegalArgumentException here, if the timeout is invalid
+      if (!IllegalArgumentException.class.getName().equals(e.getOriginalExceptionClass())) {
+        LOG.trace("Expecting only {} as the original exception class but found {}",
+                  IllegalArgumentException.class.getName(), e.getOriginalExceptionClass());
+        throw e;
+      }
+      throw new IllegalArgumentException(e.getMessage());
     } catch (TException e) {
       isValid.set(false);
       throw e;
--- a/tephra-core/src/main/java/org/apache/tephra/distributed/thrift/TBoolean.java
+++ b/tephra-core/src/main/java/org/apache/tephra/distributed/thrift/TBoolean.java
@@ -24,19 +24,29 @@
  */
 package org.apache.tephra.distributed.thrift;
 
-import org.apache.thrift.EncodingUtils;
-import org.apache.thrift.protocol.TTupleProtocol;
 import org.apache.thrift.scheme.IScheme;
 import org.apache.thrift.scheme.SchemeFactory;
 import org.apache.thrift.scheme.StandardScheme;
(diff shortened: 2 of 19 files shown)


anew:
Please do not review Thrift classes :) 
tephra-core/src/main/java/org/apache/tephra/distributed/thrift/T*.java

tephra-core/src/main/java/org/apache/tephra/distributed/thrift/TBoolean.java
```diff
@@ -24,19 +24,29 @@
  */
 package org.apache.tephra.distributed.thrift;
 
-import org.apache.thrift.EncodingUtils;
```
anew: Note that thrift does not generate imports in the order that the checkstyle rules expect. The previous version of the generated thrift stubs was apparently manually changed to reflect the correct order and to have a copyright. For the copyright, I added an instruction how to add that to the README file, but for import order, we really should not fix that manually every time we run thrift. So we see a lot of diffs in the import section here.

tephra-core/src/main/java/org/apache/tephra/distributed/TransactionServiceThriftClient.java
```diff
@@ -112,7 +113,20 @@ public Transaction startShort() throws TException {
...
+    } catch (TGenericException e) {
+      isValid.set(false);
+      // currently, we only expect IllegalArgumentException here, if the timeout is invalid
+      IllegalArgumentException actual;
+      try {
+        actual = (IllegalArgumentException)
+          Class.forName(e.getOriginalExceptionClass())
+            .getConstructor(String.class)
+            .newInstance(e.getMessage());
+      } catch (Throwable t) {
```
poornachandra: Would be good to log `t` in trace over here.

tephra-core/src/test/java/org/apache/tephra/TransactionSystemTest.java
```diff
@@ -33,25 +33,33 @@
...
 
-  // Unfortunately, in-memory mode and thrift mode throw different exceptions here
-  @Test(expected = Exception.class)
+  @Test
   public void testNegativeTimeout() throws Exception {
-    getClient().startShort(-1);
+    try {
+      getClient().startShort(-1);
+      Assert.fail("Expected illegal argument for negative timeout");
+    } catch (IllegalArgumentException e) {
```
poornachandra: What if some other exception is thrown? Would be good to use `@Test(expected = IllegalArgumentException.class)`

tephra-core/src/test/java/org/apache/tephra/TransactionSystemTest.java
```diff
@@ -33,25 +33,33 @@
...
 
-  // Unfortunately, in-memory mode and thrift mode throw different exceptions here
-  @Test(expected = Exception.class)
+  @Test
   public void testExcessiveTimeout() throws Exception {
-    getClient().startShort((int) TimeUnit.DAYS.toSeconds(10));
+    try {
+      getClient().startShort((int) TimeUnit.DAYS.toSeconds(10));
+      Assert.fail("Expected illegal argument for excessive timeout");
+    } catch (IllegalArgumentException e) {
```
poornachandra: Same here

tephra-core/src/test/java/org/apache/tephra/ThriftTransactionSystemTest.java
```diff
@@ -124,4 +127,39 @@ protected TransactionSystemClient getClient() throws Exception {
   protected TransactionStateStorage getStateStorage() throws Exception {
     return storage;
   }
+
+  @Override
+  public void testNegativeTimeout() throws Exception {
+    CountingRetryStrategyProvider.retries.set(0);
+    super.testNegativeTimeout();
+    Assert.assertEquals(0, CountingRetryStrategyProvider.retries.get());
```
poornachandra: It would be good to add a comment on why the retries should be zero here.

tephra-core/src/test/java/org/apache/tephra/TransactionSystemTest.java
anew: This is because in ThriftTransactionSystemTest, we want to assert additionally that the number of retries is 0. That cannot be done if the test case has (expected=IllegalArgumentException.class).

tephra-core/src/test/java/org/apache/tephra/TransactionSystemTest.java
anew: same as above

tephra-core/src/test/java/org/apache/tephra/ThriftTransactionSystemTest.java
anew: good point, added.

tephra-core/src/main/java/org/apache/tephra/distributed/TransactionServiceThriftClient.java
anew: Because we do not expect anything else than IllegalArgumentException? For now, I will change the code to be more readable and expect exactly that.

tephra-core/src/test/java/org/apache/tephra/TransactionSystemTest.java
poornachandra: In that case it would be good to catch the `IllegalArgumentException` in `ThriftTransactionSystemTest`. This will reduce the coupling between the tests.

anew: I don't think you can. If I annotate the base class method with (expected=...), then that is expected also for the sub class method. By the way, I don't think there is a problem with coupling between the tests. They are testing the same thing on different implementations, only that one of them wants to make an additional assertion

anew:
@poornachandra I have addressed your comments, please take a look.

tephra-core/src/test/java/org/apache/tephra/TransactionSystemTest.java
poornachandra: Sounds good. I incorrectly assumed that non-IllegalArgumentException was being ignored.

poornachandra:
LGTM

PR #19, opened 2016-11-02
TEPHRA-35 Save compaction state for pruning invalid list

poornachandra:
JIRA - https://issues.apache.org/jira/browse/TEPHRA-35

Adds ability to save prune upper bound from the transaction snapshot used for compaction.

Note that the first two commits are re-factoring existing tests.

### DIFF
--- a/tephra-core/src/main/java/org/apache/tephra/TxConstants.java
+++ b/tephra-core/src/main/java/org/apache/tephra/TxConstants.java
@@ -345,4 +345,14 @@ public class TxConstants {
     public static final byte CURRENT_VERSION = 3;
   }
 
+  /**
+   * Configuration for data janitor
+   */
+  public static final class DataJanitor {
+    public static final String PRUNE_ENABLE = "data.tx.prune.enable";
+    public static final String PRUNE_STATE_TABLE = "data.tx.prune.state.table";
+
+    public static final boolean DEFAULT_PRUNE_ENABLE = false;
+    public static final String DEFAULT_PRUNE_STATE_TABLE = "data_tx_janitor_state";
+  }
 }
--- a/tephra-core/src/main/java/org/apache/tephra/util/TxUtils.java
+++ b/tephra-core/src/main/java/org/apache/tephra/util/TxUtils.java
@@ -149,4 +149,21 @@ public class TxUtils {
   public static boolean isPreExistingVersion(long version) {
     return version < MAX_NON_TX_TIMESTAMP;
   }
+
+  /**
+   * Returns the maximum transaction that can be removed from the invalid list for the state represented by the given
+   * transaction.
+   */
+  public static long getPruneUpperBound(Transaction tx) {
+    // If there are no invalid transactions, and no in-progress transactions then we can prune the invalid list
+    // up to the current read pointer
+    if (tx.getInvalids().length == 0 && tx.getInProgress().length == 0) {
+      return tx.getReadPointer() - 1;
+    }
+
+    long maxInvalidTx =
+      tx.getInvalids().length > 0 ? tx.getInvalids()[tx.getInvalids().length - 1] : Transaction.NO_TX_IN_PROGRESS;
+    long firstInProgress = tx.getFirstInProgress();
+    return Math.min(maxInvalidTx, firstInProgress - 1);
+  }
 }
--- a/tephra-hbase-compat-1.1-base/pom.xml
+++ b/tephra-hbase-compat-1.1-base/pom.xml
@@ -28,6 +28,11 @@
   <artifactId>tephra-hbase-compat-1.1-base</artifactId>
   <name>Apache Tephra HBase 1.1 Compatibility Base</name>
 
+  <properties>
+    <hadoop.version>2.5.1</hadoop.version>
+    <hbase11.version>1.1.1</hbase11.version>
+  </properties>
+
   <packaging>pom</packaging>
   <modules>
     <module>tephra-hbase-compat-1.2-cdh</module>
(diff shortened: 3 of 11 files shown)


tephra-core/src/main/java/org/apache/tephra/TxConstants.java
```diff
@@ -345,4 +345,14 @@
...
 
+  /**
+   * Configuration for data janitor
+   */
+  public static final class DataJanitor {
+    public static final String PRUNE_ENABLE = "data.tx.prune.enable";
+    public static final String PRUNE_STATE_TABLE = "data.tx.prune.state.table";
+
+    public static final boolean DEFAULT_PRUNE_ENABLE = false;
+    public static final String DEFAULT_PRUNE_STATE_TABLE = "default:data_tx_janitor_state";
```
anew: I am not sure whether all supported HBase versions support namespaces... this assumes we have a default namespace?

tephra-core/src/main/java/org/apache/tephra/util/TxUtils.java
```diff
@@ -149,4 +149,15 @@ private static long getMaxTTL(Map<byte[], Long> ttlByFamily) {
...
+
+  /**
+   * Returns the maximum transaction that can be removed from the invalid list for the state represented by the given
+   * transaction.
+   */
+  public static long getPruneUpperBound(Transaction tx) {
+    long maxInvalidTx =
+      tx.getInvalids().length > 0 ? tx.getInvalids()[tx.getInvalids().length - 1] : Transaction.NO_TX_IN_PROGRESS;
+    long firstInProgress = tx.getFirstInProgress();
+    return Math.min(maxInvalidTx, firstInProgress - 1);
```
anew: if there is no invalid tx, and also no in-progress, the upper bound would still be the current tx id? That is, return tx.getTransactionId()-1?

tephra-hbase-compat-1.1-base/pom.xml
```diff
@@ -28,6 +28,11 @@
   <artifactId>tephra-hbase-compat-1.1-base</artifactId>
   <name>Apache Tephra HBase 1.1 Compatibility Base</name>
 
+  <properties>
```
anew: unrelated change?

tephra-hbase-compat-1.1-base/src/test/java/org/apache/tephra/hbase/InvalidListPruneTest.java
```diff
@@ -0,0 +1,203 @@
...
+          return testUtil.getConnection().getTable(pruneStateTable);
+        }
+      });
+
+    // No prune upper bound initially
+    Assert.assertEquals(-1, dataJanitorState.getPruneUpperBound(getRegionName(dataTable, Bytes.toBytes(0))));
+
+    // Create a new transaction snapshot
+    transactionSnapshot = new TransactionSnapshot(100, 100, 100, ImmutableSet.of(50L),
+                                                  ImmutableSortedMap.<Long, TransactionManager.InProgressTx>of());
```
anew: I am not sure how this works... why does creating a snapshot change the transaction state in the coprocessor?

anew:
Question: Suppose I start a transaction, which times out, and therefore goes into the invalid list. A little later HBase performs a major compaction. This transaction and all its writes are removed from the table by the DataJanitor. A little later TxManager prunes its invalid transactions, and because this tx has been removed from HBase, it removes it from the invalid list. 

The problem is if the program that started the transaction is still running. What if it performs another write after the transaction pruning? This would be an invalid version, but now it has been pruned from the invalid list and becomes visible. 

Isn't that a problem?

tephra-core/src/main/java/org/apache/tephra/TxConstants.java
poornachandra: Good point, I removed the namespace from the default value. It can always be overridden in the configuration

tephra-core/src/main/java/org/apache/tephra/util/TxUtils.java
poornachandra: Good catch, updated the code to use the current read pointer in such a case

tephra-hbase-compat-1.1-base/pom.xml
poornachandra: Yes - it is required to run the tests through IDE

poornachandra:
@anew One way to handle the above issue is to define a time limit that defines a maximum duration a transaction can be used for writing. While doing data writes, we could add some checks in the co-processor to enforce this time limit. While pruning we can use this time limit to remove invalid transactions that are older than this time limit.

anew:
@poornachandra are you planning to add this maximum duration as part of this implementation? Or is that a future improvement?

tephra-hbase-compat-1.1-base/src/test/java/org/apache/tephra/hbase/InvalidListPruneTest.java
```diff
@@ -185,6 +186,12 @@ public TransactionStateCache get() {
    */
   @SuppressWarnings("WeakerAccess")
   public static class InMemoryTransactionStateCache extends TransactionStateCache {
+    private static TransactionVisibilityState transactionSnapshot;
+
+    public static void setTransactionSnapshot(TransactionVisibilityState transactionSnapshot) {
+      InMemoryTransactionStateCache.transactionSnapshot = transactionSnapshot;
+    }
```
anew: much nicer 👍

poornachandra:
@anew I'll add the maximum duration check as a separate PR. I'll file a JIRA for it.

PR #20, opened 2016-11-22
Compute global prune upper bound using compaction state of every region

poornachandra:
JIRA - https://issues.apache.org/jira/browse/TEPHRA-198

This PR uses the compaction state recorded by co-processors in https://github.com/apache/incubator-tephra/pull/19 to compute the global prune upper bound.

TODO: add some more test cases to test DefaultDataJanitorPlugin.


tephra-hbase-compat-1.1-base/src/main/java/org/apache/tephra/hbase/coprocessor/janitor/DefaultDataJanitorPlugin.java
```diff
@@ -0,0 +1,189 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.SortedSet;
+import java.util.TreeSet;
+
+/**
+ * Default implementation of the {@link DataJanitorPlugin} for HBase
+ */
+@SuppressWarnings("WeakerAccess")
+public class DefaultDataJanitorPlugin implements DataJanitorPlugin {
```
anew: I think this is the HBaseJanitorPlugin? And if you had another store, it would be entirely different?

tephra-core/src/main/java/org/apache/tephra/janitor/DataJanitorPlugin.java
```diff
@@ -0,0 +1,59 @@
...
+ * Data janitor interface to manage the invalid transaction list.
+ * There will be one such plugin per data store that will be invoked periodically
+ * to fetch the prune upper bound for each data store.
+ * Invalid transaction list will pruned up to the minimum of prune upper bounds returned by all the plugins.
+ */
+public interface DataJanitorPlugin {
+  /**
+   * Called once at the beginning to initialize the plugin
+   */
+  void initialize(Configuration conf) throws IOException;
```
anew: at the beginning of what? The lifetime of a transaction manager? Or the beginning of a prune operation?

tephra-core/src/main/java/org/apache/tephra/janitor/DataJanitorPlugin.java
```diff
@@ -0,0 +1,59 @@
...
+
+package org.apache.tephra.janitor;
+
+import org.apache.hadoop.conf.Configuration;
+
+import java.io.IOException;
+
+/**
+ * Data janitor interface to manage the invalid transaction list.
+ * There will be one such plugin per data store that will be invoked periodically
```
anew: It would make sense to mention here that this a plugin for the tx manager. Its name is a little misleading because the DataJanitor is a coprocessor that runs in each region server.

tephra-core/src/main/java/org/apache/tephra/janitor/DataJanitorPlugin.java
```diff
@@ -0,0 +1,59 @@
...
+  /**
+   * Called once at the beginning to initialize the plugin
+   */
+  void initialize(Configuration conf) throws IOException;
+
+  /**
+   * Called periodically to fetch prune upper bound for a data store
+   *
+   * @param time start time of this prune iteration
+   * @param pruneUpperBoundForTime upper bound for prune tx id for the given start time
```
anew: this is confusing. I thought this _returns_ the prune upper bound? Why does it need the upper bound passed in as a parameter. I think you need to explain better what this parameter means.

tephra-core/src/main/java/org/apache/tephra/janitor/DataJanitorPlugin.java
```diff
@@ -0,0 +1,59 @@
...
+   */
+  long fetchPruneUpperBound(long time, long pruneUpperBoundForTime) throws IOException;
+
+  /**
+   * Called after pruning the invalid list.
+   * The plugin can use the pruneUpperBound passed to clean up its state
```
anew: is this the same pruneUpperBound that was returned by fetchPruneUpperBound? What is the semantic of this - does it mean it is guaranteed that all invalid transactions less than that upper bound have been removed from the invalid list?

tephra-hbase-compat-1.1-base/src/main/java/org/apache/tephra/hbase/coprocessor/janitor/DataJanitorState.java
```diff
@@ -19,20 +19,46 @@
...
 import java.io.IOException;
+import java.util.Map;
+import java.util.Set;
+import java.util.SortedSet;
+import java.util.TreeMap;
+import java.util.TreeSet;
+import javax.annotation.Nullable;
 
 /**
  * Persist data janitor state into an HBase table.
```
anew: Is it true that this class is shared between the DataJanitor coprocessor (for saving prune state) and the DataJanitorPlugin in the tx manager (for reading the state)? If so, then this deserves mentioning in the javadocs

tephra-hbase-compat-1.1-base/src/main/java/org/apache/tephra/hbase/coprocessor/janitor/DataJanitorState.java
```diff
@@ -58,10 +105,237 @@ public long getPruneUpperBound(byte[] regionId) throws IOException {
...
+  }
+
+  // ---------------------------------------------------
+  // ------- Methods for regions at a given time -------
+  // Key: 0x2<time><region-id>
+  // Col 't': <empty byte array>
+  // ---------------------------------------------------
+
+  /**
+   * Persist the regions for a given time
```
anew: what exactly does this persist? Simply what regions existed at that time?

tephra-hbase-compat-1.1-base/src/main/java/org/apache/tephra/hbase/coprocessor/janitor/DefaultDataJanitorPlugin.java
```diff
@@ -0,0 +1,189 @@
...
+  public long fetchPruneUpperBound(long time, long pruneUpperBoundForTime) throws IOException {
+    LOG.debug("Fetching prune upper bound for time {} and max prune upper bound {}", time, pruneUpperBoundForTime);
+    if (time < 0 || pruneUpperBoundForTime < 0) {
+      return -1;
+    }
+
+    SortedSet<byte[]> transactionalRegions = getTransactionalRegions();
+    if (!transactionalRegions.isEmpty()) {
+      LOG.debug("Saving {} transactional regions for time {}", transactionalRegions.size(), time);
+      dataJanitorState.saveRegionsForTime(time, transactionalRegions);
```
anew: I am not sure I understand. Why does fetch() have to save anything? Shouldn't it just read (=fetch)?

tephra-hbase-compat-1.1-base/src/main/java/org/apache/tephra/hbase/coprocessor/janitor/DefaultDataJanitorPlugin.java
```diff
@@ -0,0 +1,189 @@
...
+          } else {
+            LOG.debug("{} is not a transactional table", tableDescriptor.getTableName());
+          }
+        }
+      }
+    }
+    return regions;
+  }
+
+  private long computePruneUpperBound(TimeRegions timeRegions) throws IOException {
```
anew: It would really help to explain this method in comments.

tephra-hbase-compat-1.1-base/src/main/java/org/apache/tephra/hbase/coprocessor/janitor/DataJanitorState.java
poornachandra: Added javadoc

tephra-core/src/main/java/org/apache/tephra/janitor/TransactionPruningPlugin.java
```diff
@@ -0,0 +1,90 @@
...
+  /**
+   * Called once when the Transaction Service starts up.
+   *
+   * @param conf configuration for the plugin
+   */
+  void initialize(Configuration conf) throws IOException;
+
+  /**
+   * Called periodically to fetch prune upper bound for a data store. The plugin examines the state of data cleanup
+   * in the data store and determines the smallest invalid transaction whose writes no longer exist in the data
```
anew: not sure if this is correct. I think the intention is: It is the greatest upper bound that guarantees that the writes of all invalid transaction less than this bound no longer exist.

anew: or a greatest lower bound for transaction ids that may not be pruned?

tephra-hbase-compat-1.1-base/src/main/java/org/apache/tephra/hbase/coprocessor/janitor/HBaseTransactionPruningPlugin.java
```diff
@@ -0,0 +1,289 @@
...
+
+      Map<byte[], Long> pruneUpperBoundRegions = dataJanitorState.getPruneUpperBoundForRegions(transactionalRegions);
+      logPruneUpperBoundRegions(pruneUpperBoundRegions);
+      // If prune upper bounds are found for all the transactional regions, then compute the prune upper bound
+      // across all regions
+      if (!transactionalRegions.isEmpty() && pruneUpperBoundRegions.size() == transactionalRegions.size()) {
+        long pruneUpperBoundForTime = dataJanitorState.getPruneUpperBoundForTime(time);
+        LOG.debug("Found max prune upper bound {} for time {}", pruneUpperBoundForTime, time);
+        // If pruneUpperBoundForTime is not recorded then that means the data is not complete for these regions
+        if (pruneUpperBoundForTime != -1) {
```
anew: It would be good to debug-log if it is -1. That would indicate the reason for the debug message in line 264.

tephra-hbase-compat-1.1-base/src/main/java/org/apache/tephra/hbase/coprocessor/janitor/HBaseTransactionPruningPlugin.java
```diff
@@ -0,0 +1,289 @@
...
+  public void pruneComplete(long time, long maxPrunedInvalid) throws IOException {
+    LOG.debug("Prune complete for time {} and prune upper bound {}", time, maxPrunedInvalid);
+    if (time < 0 || maxPrunedInvalid < 0) {
+      return;
+    }
+
+    // Get regions for given time, so as to not delete them
+    TimeRegions regionsToExclude = dataJanitorState.getRegionsOnOrBeforeTime(time);
+    if (regionsToExclude != null) {
+      LOG.debug("Deleting stale region - prune upper bound record before {}", maxPrunedInvalid);
```
anew: regions with p.u.b. recorded before

anew: So this will only delete the entries for regions that do not exist any longer? Meaning, for regions that still exist, we never clean up that state? It's not clear why we have to keep it.

tephra-hbase-compat-1.1-base/src/main/java/org/apache/tephra/hbase/coprocessor/janitor/HBaseTransactionPruningPlugin.java
```diff
@@ -0,0 +1,289 @@
...
+      dataJanitorState.deleteRegionsWithPruneUpperBoundBefore(maxPrunedInvalid, regionsToExclude.getRegions());
+    } else {
+      LOG.warn("Cannot find saved regions on or before time {}", time);
+    }
+    long pruneTime = maxPrunedInvalid / TxConstants.MAX_TX_PER_MS;
```
anew: It seems that there should be a helper method to extract the time stamp from a transaction id?

tephra-core/src/main/java/org/apache/tephra/janitor/TransactionPruningPlugin.java
```diff
@@ -0,0 +1,90 @@
...
+   * store. It then returns this smallest invalid transaction as the prune upper bound for this data store.
+   *
+   * @param time start time of this prune iteration in milliseconds
+   * @param pruneUpperBoundForTime the largest invalid transaction that can be possibly removed
+   *                               from the invalid list for the given time.
+   *                               In terms of HBase, this is the smallest not in-progress transaction that will
+   *                               not have writes in any HBase regions that are created after the given time.
+   *                               The plugin will typically return a reduced upper bound based on the state of
+   *                               the invalid transaction data clean up in the data store.
```
poornachandra: You are right, I removed the HBase part from here since it is confusing and updated the javadoc.

tephra-hbase-compat-1.1-base/src/main/java/org/apache/tephra/hbase/coprocessor/janitor/HBaseTransactionPruningPlugin.java
```diff
@@ -179,15 +180,17 @@ public void pruneComplete(long time, long maxPrunedInvalid) throws IOException {
       return;
     }
 
-    // Get regions for given time, so as to not delete them
+    // Get regions for the given time, so as to not delete them. The prune upper bounds for regions are recorded
+    // by TransactionProcessor and the deletion is done by this class. To avoid update/delete race condition,
+    // we only delete stale regions.
```
anew: OK, the comment makes this clear that it is because of race condition. Not sure what kind of race conditions? Also, I believe my comment still holds that this means, as long as a region exists, we will never delete its entries here. Do we really have to keep a record of a region's existence for every single pruneInterval since the beginning of time? That can be a lot of records over time.

poornachandra:
Thanks for the review @anew. I've addressed the comments, please take another look.

> Regarding correctness, my major concern is that the (min(max invalid), min(inProgress)-1) is not a safe bound for transaction that are guaranteed not to write any more.
I have filed TEPHRA-199 to address this, I'll make the changes in the next PR.

Do you have a suggestion for a better name for `pruneUpperBoundTime`? Basically this is an upper bound for transactions that are not active (i.e, can be still used for writing).

tephra-hbase-compat-1.1-base/src/main/java/org/apache/tephra/hbase/coprocessor/janitor/HBaseTransactionPruningPlugin.java
```diff
@@ -259,6 +262,8 @@ private long computePruneUpperBound(TimeRegions timeRegions) throws IOException
         if (pruneUpperBoundForTime != -1) {
           Long minPruneUpperBoundRegions = Collections.min(pruneUpperBoundRegions.values());
           return Math.min(pruneUpperBoundForTime, minPruneUpperBoundRegions);
+        } else {
+          LOG.debug("Ignoring invalid prune upper bound -1 for time {}", time);
```
anew: A better message would be: Ignoring regions for time {} because no pruneUpperBound was found for that time and the data must be incomplete.

tephra-hbase-compat-1.1-base/src/main/java/org/apache/tephra/hbase/coprocessor/janitor/HBaseTransactionPruningPlugin.java
```diff
@@ -259,6 +262,8 @@ private long computePruneUpperBound(TimeRegions timeRegions) throws IOException
...
         }
       } else {
         if (LOG.isDebugEnabled()) {
```
anew: Maybe the message here should also be rephrased: "Ignoring regions for time {} because not all of the regions recorded a pruneUpperBound."

This may seem picky, but remember that these messages will be seen by devops who does not understand the internals. If you rephrase it this way, they will understand better what is gong on.

tephra-hbase-compat-1.1-base/src/main/java/org/apache/tephra/hbase/coprocessor/janitor/HBaseTransactionPruningPlugin.java
```diff
@@ -0,0 +1,289 @@
...
+ *
+ * <p/>
+ * Above, when we find <i>(t1, set of regions)</i>, there may a region that was created after time <i>t1</i>,
+ * but has a data write from an invalid transaction that is smaller than the prune upper bounds of all
+ * regions in <i>(t1, set of regions)</i>. This is possible because <i>(region, prune upper bound)</i> persisted by
+ * TransactionProcessor is always the latest prune upper bound for a region.
+ * <br/>
+ * However a region created after time <i>t1</i> cannot have writes from an invalid transaction that is smaller than
+ * <i>min(max(invalid list), min(in-progress list) - 1)</i> at the time the region was created.
+ * Since we limit the plugin prune upper bound using <i>(t1, prune upper bound)</i>, there should be no invalid
```
anew: Perhaps we should not document the min(max(invalid list), min(in-progress list) - 1) here. This class actually does not compute that, it is the caller that does. When we fix TEPHRA-199, then it will become obsolete anyway.

anew:
I did another pass. Only a few more comments...

tephra-hbase-compat-1.1-base/src/main/java/org/apache/tephra/hbase/coprocessor/janitor/HBaseTransactionPruningPlugin.java
poornachandra: Sure, will update the message

anew:
LGTM

poornachandra:
@anew Please take a look at the last rename commit

anew:
Thanks for the rename, I think this makes the code much clearer. LGTM

PR #22, opened 2016-12-06
Transaction pruning service

poornachandra:
https://issues.apache.org/jira/browse/TEPHRA-203
https://issues.apache.org/jira/browse/TEPHRA-199

This PR has three things -
 * Define and enforce maximum transaction lifetime where transactions can be used for data writes
 * Add `TransactionPruningService` to periodically schedule transaction prune run
* Rename package name janitor to txprune

It would be easier to review the commits independently.

### DIFF
--- a/tephra-core/src/main/java/org/apache/tephra/TxConstants.java
+++ b/tephra-core/src/main/java/org/apache/tephra/TxConstants.java
@@ -354,13 +354,35 @@ public class TxConstants {
   }
 
   /**
-   * Configuration for data janitor
+   * Configuration for invalid transaction pruning
    */
-  public static final class DataJanitor {
+  public static final class TransactionPruning {
+    /**
+     * Flag to enable automatic invalid transaction pruning.
+     */
     public static final String PRUNE_ENABLE = "data.tx.prune.enable";
+    /**
+     * The table used to store intermediate state when pruning is enabled.
+     */
     public static final String PRUNE_STATE_TABLE = "data.tx.prune.state.table";
+    /**
+     * Interval in seconds to schedule prune run.
+     */
+    public static final String PRUNE_INTERVAL = "data.tx.prune.interval";
+    /**
+     * Comma separated list of invalid transaction pruning plugins to load
+     */
+    public static final String PLUGINS = "data.tx.prune.plugins";
+    /**
+     * Class name for the plugins will be plugin-name + ".class" suffix
+     */
+    public static final String PLUGIN_CLASS_SUFFIX = ".class";
 
     public static final boolean DEFAULT_PRUNE_ENABLE = false;
     public static final String DEFAULT_PRUNE_STATE_TABLE = "data_tx_janitor_state";
+    public static final long DEFAULT_PRUNE_INTERVAL = TimeUnit.HOURS.toSeconds(6);
+    public static final String DEFAULT_PLUGIN = "data.tx.prune.plugin.default";
+    public static final String DEFAULT_PLUGIN_CLASS =
+      "org.apache.tephra.hbase.txprune.HBaseTransactionPruningPlugin";
   }
 }
--- a/tephra-core/src/main/java/org/apache/tephra/TxConstants.java
+++ b/tephra-core/src/main/java/org/apache/tephra/TxConstants.java
@@ -158,6 +158,14 @@ public class TxConstants {
      * The default value for the transaction timeout limit, in seconds: unlimited.
      */
     public static final int DEFAULT_TX_MAX_TIMEOUT = Integer.MAX_VALUE;
+    /**
+     * The maximum time in seconds that a transaction can be used for data writes.
+     */
+    public static final String CFG_TX_MAX_LIFETIME = "data.tx.max.lifetime";
+    /**
+     * The default value for the maximum transaction lifetime.
+     */
+    public static final int DEFAULT_TX_MAX_LIFETIME = (int) TimeUnit.HOURS.toSeconds(25);
     /** The frequency (in seconds) to perform periodic snapshots, or 0 for no periodic snapshots. */
     public static final String CFG_TX_SNAPSHOT_INTERVAL = "data.tx.snapshot.interval";
     /** Default value for frequency of periodic snapshots of transaction state. */
--- a/tephra-core/src/main/java/org/apache/tephra/txprune/TransactionPruningRunnable.java
+++ b/tephra-core/src/main/java/org/apache/tephra/txprune/TransactionPruningRunnable.java
@@ -1,0 +1,128 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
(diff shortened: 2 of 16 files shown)


tephra-core/src/main/java/org/apache/tephra/TxConstants.java
```diff
@@ -158,6 +158,14 @@
      * The default value for the transaction timeout limit, in seconds: unlimited.
      */
     public static final int DEFAULT_TX_MAX_TIMEOUT = Integer.MAX_VALUE;
+    /**
+     * The maximum time in seconds that a transaction can be used for data writes.
+     */
+    public static final String CFG_TX_MAX_LIFETIME = "data.tx.max.lifetime";
```
gokulavasan: This is only for short transactions right? Wasn't very clear from the comment.

gokulavasan: Nevermind, it looks like it is for all transactions.

tephra-core/src/main/java/org/apache/tephra/TxConstants.java
```diff
@@ -158,6 +158,14 @@
      * The default value for the transaction timeout limit, in seconds: unlimited.
      */
     public static final int DEFAULT_TX_MAX_TIMEOUT = Integer.MAX_VALUE;
+    /**
+     * The maximum time in seconds that a transaction can be used for data writes.
```
anew: Does this apply to all transaction, including long-running ones? I think it helps to say that after this time, all writes to the table will fail, and that is ensured by Tephra's coprocessors.

You should also document that this value influences invalid tx pruning.

tephra-hbase-compat-1.1-base/src/main/java/org/apache/tephra/hbase/TransactionAwareHTable.java
```diff
@@ -649,6 +649,7 @@ private Delete transactionalizeAction(Delete delete) throws IOException {
         txDelete.setAttribute(entry.getKey(), entry.getValue());
     }
     txDelete.setDurability(delete.getDurability());
+    addToOperation(txDelete, tx);
```
anew: unrelated change?

tephra-hbase-compat-1.1-base/src/main/java/org/apache/tephra/hbase/coprocessor/TransactionProcessor.java
```diff
@@ -179,6 +187,13 @@ public void preGetOp(ObserverContext<RegionCoprocessorEnvironment> e, Get get, L
   }
 
   @Override
+  public void prePut(ObserverContext<RegionCoprocessorEnvironment> e, Put put, WALEdit edit, Durability durability)
+    throws IOException {
+    Transaction tx = getFromOperation(put);
+    ensureValidTxLifetime(tx);
+  }
+
+  @Override
```
anew: what about increment operations? and readless increments?

tephra-core/src/main/java/org/apache/tephra/TxConstants.java
poornachandra: Yes, this limit applies to all transactions. Will update the documentation.

tephra-hbase-compat-1.1-base/src/main/java/org/apache/tephra/hbase/coprocessor/TransactionProcessor.java
poornachandra: Increment and other operations are non-transactional today. Tephra does not support readless increments.

anew: oh i that case it is ok.

tephra-hbase-compat-1.1-base/src/main/java/org/apache/tephra/hbase/TransactionAwareHTable.java
anew: I see

tephra-core/src/main/java/org/apache/tephra/txprune/TransactionPruningRunnable.java
```diff
@@ -0,0 +1,122 @@
...
+        LOG.info("Not pruning invalid list since minimum prune upper bound ({}) is less than 1", minPruneUpperBound);
+        return;
+      }
+
+      long[] invalids = tx.getInvalids();
+      TreeSet<Long> toTruncate = new TreeSet<>();
+      LOG.debug("Invalid list: {}", invalids);
+      for (long invalid : invalids) {
+        if (invalid <= minPruneUpperBound) {
+          toTruncate.add(invalid);
```
anew: do we have to compute this list? Or can the txManager simply remove all transactions that are older than the minPruneUpperBound?

tephra-core/src/main/java/org/apache/tephra/util/TxUtils.java
```diff
@@ -163,8 +167,22 @@ public static long getPruneUpperBound(Transaction tx) {
...
+    // transaction until it exhausts its max lifetime.
+    long elapsedTime = currentTimeMillis - TxUtils.getTimestamp(maxInvalidTx);
+    long invalidTxBound;
+    if (elapsedTime > txMaxLifetimeMillis) {
+      // maxInvalidTx is past its max lifetime
+      invalidTxBound = maxInvalidTx;
+    } else {
+      // Reduce the lifetime of maxInvalidTx, so that it cannot be used for any more writes
+      long remainingTime = txMaxLifetimeMillis - elapsedTime;
+      invalidTxBound = maxInvalidTx - remainingTime * TxConstants.MAX_TX_PER_MS - 1;
```
anew: Independent of maxInvalidTx, isn't this always the same as (currentTimeMillis - txMaxLifeTimeMillis) * MAX_TX_PER_MS?

poornachandra: Good point. Now that I think about it, we can further simplify the method `TxUtils.getPruneUpperBound()` to simply return `(currentTimeMillis - txMaxLifetimeMillis) * TxConstants.MAX_TX_PER_MS - 1`, irrespective of whether there are in-progress transactions or not. Since no transaction (in-progress or invalid) can be used for writes after `txMaxLifetimeMillis`.

What do you say?

tephra-core/src/main/java/org/apache/tephra/txprune/TransactionPruningRunnable.java
poornachandra: The reason to build the list is so that we can use the bulk `removeAll()` method of invalid list, rather than removing the invalid ids one by one.

tephra-core/src/main/java/org/apache/tephra/util/TxUtils.java
anew: agreed.

poornachandra: On further thought, we will need to compute inactive transaction bound separately from prune upper bound. 

Inactive transaction bound can be computed using just current time and tx max lifetime as ` (currentTimeMillis - txMaxLifetimeMillis) * TxConstants.MAX_TX_PER_MS - 1`. 

However, prune upper bound is a record of what invalid data was removed during a major compaction, and hence will need to use the invalid list and in-progress list from the transaction snapshot used at the time of the major compaction.

anew: yes, that comes from each of the regions, right?

poornachandra: Yes

poornachandra:
@anew I have separated out the computation of prune upper bound and inactive tx bound as discussed, please take another look.

anew:
LGTM

PR #23, opened 2016-12-08
[Tephra-201] Store checkpoints in in-progress list 

anew:
See https://issues.apache.org/jira/browse/TEPHRA-201 for the design proposal implemented by this PR.

### DIFF
--- a/tephra-core/src/main/java/org/apache/tephra/TransactionManager.java
+++ b/tephra-core/src/main/java/org/apache/tephra/TransactionManager.java
@@ -340,35 +340,37 @@ public class TransactionManager extends AbstractService {
 
   private void cleanupTimedOutTransactions() {
     List<TransactionEdit> invalidEdits = null;
-    this.logReadLock.lock();
+    logReadLock.lock();
     try {
       synchronized (this) {
         if (!isRunning()) {
           return;
         }
-
         long currentTime = System.currentTimeMillis();
-        List<Long> timedOut = Lists.newArrayList();
+        Map<Long, InProgressType> timedOut = Maps.newHashMap();
         for (Map.Entry<Long, InProgressTx> tx : inProgress.entrySet()) {
           long expiration = tx.getValue().getExpiration();
           if (expiration >= 0L && currentTime > expiration) {
             // timed out, remember tx id (can't remove while iterating over entries)
-            timedOut.add(tx.getKey());
+            timedOut.put(tx.getKey(), tx.getValue().getType());
             LOG.info("Tx invalid list: added tx {} because of timeout", tx.getKey());
           } else if (expiration < 0) {
             LOG.warn("Transaction {} has negative expiration time {}. Likely cause is the transaction was not " +
                        "migrated correctly, this transaction will be expired immediately",
                      tx.getKey(), expiration);
-            timedOut.add(tx.getKey());
+            timedOut.put(tx.getKey(), InProgressType.LONG);
           }
         }
         if (!timedOut.isEmpty()) {
           invalidEdits = Lists.newArrayListWithCapacity(timedOut.size());
-          invalid.addAll(timedOut);
-          for (long tx : timedOut) {
-            committingChangeSets.remove(tx);
-            inProgress.remove(tx);
-            invalidEdits.add(TransactionEdit.createInvalid(tx));
+          invalid.addAll(timedOut.keySet());
+          for (Map.Entry<Long, InProgressType> tx : timedOut.entrySet()) {
+            inProgress.remove(tx.getKey());
+            // checkpoints never go into the committing change sets or the edits
+            if (!InProgressType.CHECKPOINT.equals(tx.getValue())) {
+              committingChangeSets.remove(tx.getKey());
+              invalidEdits.add(TransactionEdit.createInvalid(tx.getKey()));
+            }
           }
 
           // todo: find a more efficient way to keep this sorted. Could it just be an array?
@@ -529,13 +531,13 @@ public class TransactionManager extends AbstractService {
         // handle null expiration
         long newExpiration = getTxExpirationFromWritePointer(writePointer, defaultLongTimeout);
         InProgressTx compatTx =
-          new InProgressTx(entry.getValue().getVisibilityUpperBound(), newExpiration, TransactionType.LONG,
+          new InProgressTx(entry.getValue().getVisibilityUpperBound(), newExpiration, InProgressType.LONG,
               entry.getValue().getCheckpointWritePointers());
         entry.setValue(compatTx);
       } else if (entry.getValue().getType() == null) {
         InProgressTx compatTx =
           new InProgressTx(entry.getValue().getVisibilityUpperBound(), entry.getValue().getExpiration(),
-                           TransactionType.SHORT, entry.getValue().getCheckpointWritePointers());
+                           InProgressType.SHORT, entry.getValue().getCheckpointWritePointers());
         entry.setValue(compatTx);
       }
     }
@@ -620,7 +622,15 @@ public class TransactionManager extends AbstractService {
               if (type == null) {
                 InProgressTx inProgressTx = inProgress.get(edit.getWritePointer());
                 if (inProgressTx != null) {
-                  type = inProgressTx.getType();
+                  InProgressType inProgressType = inProgressTx.getType();
+                  if (InProgressType.CHECKPOINT.equals(inProgressType)) {
+                    // this should never happen, because checkpoints never go into the log edits;
+                    LOG.debug("Ignoring ABORTED edit for a checkpoint transaction {}", edit.getWritePointer());
+                    break;
+                  }
+                  if (inProgressType != null) {
+                    type = inProgressType.getTransactionType();
+                  }
                 } else {
                   // If transaction is not in-progress, then it has either been already aborted or invalidated.
                   // We cannot determine the transaction's state based on current information, to be safe invalidate it.
@@ -788,6 +798,11 @@ public class TransactionManager extends AbstractService {
 
   private void addInProgressAndAdvance(long writePointer, long visibilityUpperBound,
                                        long expiration, TransactionType type) {
+    addInProgressAndAdvance(writePointer, visibilityUpperBound, expiration, InProgressType.of(type));
+  }
+
+  private void addInProgressAndAdvance(long writePointer, long visibilityUpperBound,
+                                       long expiration, InProgressType type) {
     inProgress.put(writePointer, new InProgressTx(visibilityUpperBound, expiration, type));
     advanceWritePointer(writePointer);
   }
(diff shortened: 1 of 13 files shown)


tephra-core/src/main/java/org/apache/tephra/TransactionManager.java
```diff
@@ -917,6 +933,15 @@ private void doCommit(long transactionId, long writePointer, Set<ChangeId> chang
         invalidArray = invalid.toLongArray();
         LOG.info("Tx invalid list: removed committed tx {}", transactionId);
       }
+    } else {
+      long[] checkpointPointers = previous.getCheckpointWritePointers().toLongArray();
+      if (checkpointPointers != null && checkpointPointers.length > 0) {
+        // adjust the write pointer to be the last checkpoint of the tx and remove all checkpoints from inProgress
+        writePointer = checkpointPointers[checkpointPointers.length - 1];
+        for (long checkpointPointer : checkpointPointers) {
+          inProgress.remove(checkpointPointer);
```
chtyim: the `LongArrayList` returned by `getCheckpointWritePointers()` method is actually an instance of `Collection<Long>` as well, so you can just call `inProgress.keySet().removeAll(previous.getCheckpointWritePointers())`

tephra-core/src/main/java/org/apache/tephra/TransactionManager.java
```diff
@@ -917,6 +933,15 @@ private void doCommit(long transactionId, long writePointer, Set<ChangeId> chang
         invalidArray = invalid.toLongArray();
         LOG.info("Tx invalid list: removed committed tx {}", transactionId);
       }
+    } else {
+      long[] checkpointPointers = previous.getCheckpointWritePointers().toLongArray();
+      if (checkpointPointers != null && checkpointPointers.length > 0) {
+        // adjust the write pointer to be the last checkpoint of the tx and remove all checkpoints from inProgress
+        writePointer = checkpointPointers[checkpointPointers.length - 1];
```
chtyim: To avoid boxing and unboxing (assuming you are using the LongArrayList directly, without copying to an array first), you can do `checkpointWritePointers.getLong(checkpointWritePointers.size())`.

tephra-core/src/main/java/org/apache/tephra/TransactionManager.java
```diff
@@ -958,25 +983,31 @@ private void doAbort(long writePointer, long[] checkpointWritePointers, Transact
...
         LOG.info("Tx invalid list: removed aborted tx {}", writePointer);
-        // removed a tx from excludes: must move read pointer
-        moveReadPointerIfNeeded(writePointer);
       }
-    } else {
-      // removed a tx from excludes: must move read pointer
-      moveReadPointerIfNeeded(writePointer);
     }
+    if (!isInvalid) {
+      if (checkpointWritePointers != null) {
```
chtyim: These two `if`s can be combined with a `&&` to have the code less indented.

tephra-core/src/main/java/org/apache/tephra/TransactionManager.java
```diff
@@ -958,25 +983,31 @@ private void doAbort(long writePointer, long[] checkpointWritePointers, Transact
...
         invalidArray = invalid.toLongArray();
         LOG.info("Tx invalid list: removed aborted tx {}", writePointer);
-        // removed a tx from excludes: must move read pointer
-        moveReadPointerIfNeeded(writePointer);
       }
-    } else {
-      // removed a tx from excludes: must move read pointer
-      moveReadPointerIfNeeded(writePointer);
     }
+    if (!isInvalid) {
```
chtyim: I found this variable name less clear on the actual action. Probably better call it `removeInProgressCheckpoints`? Because it is actually what the condition is about. It seems less clear to me why "not invalid" means something needs to be removed from the inprogress when reading this part of the code. I have to trace back to see how this variable is being set. If the variable is "removeInProgressCheckpoints", it's more obvious that removal needs to be performed here and I can optionally trace the code if I am interested in how it is being set.

tephra-core/src/main/java/org/apache/tephra/TransactionManager.java
```diff
@@ -348,27 +348,30 @@ private void cleanupTimedOutTransactions() {
         }
 
         long currentTime = System.currentTimeMillis();
-        List<Long> timedOut = Lists.newArrayList();
+        Map<Long, InProgressType> timedOut = Maps.newHashMap();
```
anew: Agreed, but we will fix that in later PRs

tephra-core/src/main/java/org/apache/tephra/TransactionManager.java
```diff
@@ -917,6 +933,15 @@ private void doCommit(long transactionId, long writePointer, Set<ChangeId> chang
         invalidArray = invalid.toLongArray();
         LOG.info("Tx invalid list: removed committed tx {}", transactionId);
       }
+    } else {
+      long[] checkpointPointers = previous.getCheckpointWritePointers().toLongArray();
```
anew: ah yeah. I copied this from existing code and did not attempt to make it better... will do now

tephra-core/src/main/java/org/apache/tephra/TransactionManager.java
chtyim: How about the "write lock" comment? Currently we are only holding the read lock and modifying the invalid list.

tephra-core/src/main/java/org/apache/tephra/TransactionManager.java
```diff
@@ -917,6 +933,15 @@ private void doCommit(long transactionId, long writePointer, Set<ChangeId> chang
...
+      if (checkpointPointers != null && checkpointPointers.length > 0) {
```
anew: I guess not. Copied from existing code...

anew: I can't find a away that this could be null. Changed it everywhere it is used to assume it is not null

tephra-core/src/main/java/org/apache/tephra/TransactionManager.java
anew: you mean size()-1? Done

chtyim: yes :)

tephra-core/src/main/java/org/apache/tephra/TransactionManager.java
anew: good point.

tephra-core/src/main/java/org/apache/tephra/TransactionManager.java
anew: the meaning is actually whether the parent has been invalidated. If so, then the previous if() has removed all children from the invalid set. In all other cases, we have to assume they are still in the inProgress map and remove them. But ok, I can rename it.

tephra-core/src/main/java/org/apache/tephra/TransactionManager.java
anew: good point. done

tephra-core/src/main/java/org/apache/tephra/TransactionManager.java
```diff
@@ -1319,20 +1344,60 @@ public void shutdown() {
...
+  /**
    * Represents some of the info on in-progress tx
    */
   public static final class InProgressTx {
     /** the oldest in progress tx at the time of this tx start */
     private final long visibilityUpperBound;
     private final long expiration;
-    private final TransactionType type;
+    private final InProgressType type;
     private LongArrayList checkpointWritePointers = new LongArrayList();
```
anew: not introduced by this PR, but fixed now.

anew: ditto

tephra-core/src/main/java/org/apache/tephra/TransactionManager.java
chtyim: I am just not sure the correctness of that. Probably we can keep the existing way, and have another PR just focus on lock (not necessarily target for perf improvement, but rather for the correctness).

anew:
thanks for the review, @chtyim. I addressed your comments. Please take another look.

tephra-core/src/main/java/org/apache/tephra/TransactionManager.java
anew: now I already changed it...

tephra-core/src/main/java/org/apache/tephra/TransactionManager.java
```diff
@@ -340,48 +340,51 @@ public long getSleepMillis() {
...
+        } else if (expiration < 0) {
+          LOG.warn("Transaction {} has negative expiration time {}. Likely cause is the transaction was not " +
+                     "migrated correctly, this transaction will be expired immediately",
+                   tx.getKey(), expiration);
+          timedOut.put(tx.getKey(), InProgressType.LONG);
         }
-        if (!timedOut.isEmpty()) {
+      }
+      if (!timedOut.isEmpty()) {
+        logWriteLock.lock();
```
chtyim: Looking at the code again, my understand of the lock in this method is wrong. The whole method itself is synchronized on `this` object (with `synchronize(this)` covering almost the entire method). The `logReadLock` is for locking on the transaction log writer, which apparently it only needs the read lock in order to call `appendToLog`. It's probably safer to revert to old way of locking first. Need further study on the locking mechanism on this class (apparently there are three locks, one on `this` logic, one read and one write lock on the tx log).

Sorry for my previous comment.

anew: agreed. ok. will revert

tephra-core/src/main/java/org/apache/tephra/TransactionManager.java
```diff
@@ -1011,10 +1038,9 @@ private boolean doInvalidate(long writePointer) {
       } else {
         // invalidate any checkpoint write pointers
         LongArrayList childWritePointers = previous.getCheckpointWritePointers();
-        if (childWritePointers != null) {
-          for (int i = 0; i < childWritePointers.size(); i++) {
-            invalid.add(childWritePointers.get(i));
-          }
+        for (long childWritePointer : childWritePointers) {
+          invalid.add(childWritePointer);
+          inProgress.remove(childWritePointer);
```
chtyim: Missed this one in last review, we can use the `keySet().removeAll(childWritePointers)` as well (and check if `!childWritePointers.isEmpty()` first). The same for invalid, it can be `invalid.addAll(childWritePointers)`

chtyim:
Just one last comment, the rest LGTM

anew:
Tests passed. I am gonna squash and commit this.

PR #26, opened 2017-01-10
TEPHRA-208 Compaction removes rows incorrectly for family deletes wit…

twdsilva:
…h column level conflict detection

@poornachandra  @gokulavasan

### DIFF
--- a/tephra-hbase-compat-0.96/src/main/java/org/apache/tephra/hbase/coprocessor/TransactionVisibilityFilter.java
+++ b/tephra-hbase-compat-0.96/src/main/java/org/apache/tephra/hbase/coprocessor/TransactionVisibilityFilter.java
@@ -291,6 +291,7 @@ public class TransactionVisibilityFilter extends FilterBase {
 
   private static final class DeleteTracker {
     private long familyDeleteTs;
+    private byte[] rowKey;
 
     public static boolean isFamilyDelete(Cell cell) {
       return !TxUtils.isPreExistingVersion(cell.getTimestamp()) &&
@@ -300,14 +301,17 @@ public class TransactionVisibilityFilter extends FilterBase {
 
     public void addFamilyDelete(Cell delete) {
       this.familyDeleteTs = delete.getTimestamp();
+      this.rowKey = Bytes.copy(delete.getRowArray(), delete.getRowOffset(), delete.getRowLength());
     }
 
     public boolean isDeleted(Cell cell) {
-      return cell.getTimestamp() <= familyDeleteTs;
+      return rowKey != null && Bytes.compareTo(cell.getRowArray(), cell.getRowOffset(), 
+        cell.getRowLength(), rowKey, 0, rowKey.length) == 0 && cell.getTimestamp() <= familyDeleteTs;
     }
 
     public void reset() {
       this.familyDeleteTs = 0;
+      this.rowKey = null;
     }
   }
 }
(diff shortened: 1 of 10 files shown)


twdsilva:
I'm not sure why the test is failing for the hbase 1.2 cdh version. I will look into it.

twdsilva:
testFamilyDeleteWithCompaction() fails for the hbase 1.2 cdh version because HDFSTransactionStateStorage cannot be initialized as it can't find the snapshot directory in config, however this test passes in the other hbase versions. 

@poornachandra  Any idea what could be happening here?
```
2017-01-12 11:00:14,329 - ERROR [HDFSTransactionStateStorage STARTING:o.a.z.s.NIOServerCnxnFactory$1@44] - Thread Thread[HDFSTransactionStateStorage STARTING,5,main] died
java.lang.IllegalStateException: Snapshot directory is not configured.  Please set data.tx.snapshot.dir in configuration.
	at com.google.common.base.Preconditions.checkState(Preconditions.java:149) ~[guava-13.0.1.jar:na]
	at org.apache.tephra.persist.HDFSTransactionStateStorage.startUp(HDFSTransactionStateStorage.java:94) ~[tephra-core-0.11.0-incubating-SNAPSHOT.jar:0.11.0-incubating-SNAPSHOT]
	at com.google.common.util.concurrent.AbstractIdleService$1$1.run(AbstractIdleService.java:43) ~[guava-13.0.1.jar:na]
	at java.lang.Thread.run(Thread.java:745) ~[na:1.7.0_76]
```

gokulavasan:
@twdsilva Curious why we need to use HDFSTransactionStateStorage to test this? Is there any problem with running it with the existing InMemoryTransactionStateStorage?

twdsilva:
Compaction wasn't removing deleted rows when InMemoryTransactionStateStorage() was used. I think this had something to do with how visibilityUpperBound was set in IncludeInProgressFilter based on the txn snapshot.

gokulavasan:
oh yes that is right. my bad. thanks for the clarification @twdsilva

poornachandra:
@twdsilva Some code changes have not been ported over to file tephra-hbase-compat-1.1-base/src/test/java/org/apache/tephra/hbase/TransactionAwareHTableTest.java, hence the tests are failing. For instance, HDFSTransactionStateStorage is not initialized and miniDFS cluster is not started in that file.

twdsilva:
@gokulavasan , thanks.  I have fixed the issue and the tests pass now.

poornachandra:
@twdsilva The changes look good. However, we recently did a re-factoring of TransactionAwareHTableTest by moving mini cluster startup/shutdown to a base class AbstractHBaseTableTest. It would be good if non-Transaction Manager startup/shutdown code and configuration setup in this PR are moved to AbstractHBaseTableTest.

twdsilva:
@poornachandra  Thanks for the review. I have made the change.

gokulavasan:
@twdsilva Couple of comments:
i) It would be good to make changes to only  one compat-module and then open it for review and once the comments are addressed and changes look good, you can then proceed to port it to other modules and get a final lgtm before merging. That way you don't have to address changes in all compat modules everytime

ii) I think what @poornachandra meant was, we currently start the miniCluster in AbstractHBaseTableTest. And so we can move the code where you start the MiniDFSCluster to AbstractHBaseTableTest#startMiniCluster method.

twdsilva:
@gokulavasan I didn't move the MiniDFSCluster initialization into AbstractHBaseTableTest since the other tests that extend AbstractHBaseTableTest didn't require staring the MiniDFSCluster.

gokulavasan:
@poornachandra Do you have any thoughts on https://github.com/apache/incubator-tephra/pull/26#issuecomment-276127210 ?

poornachandra:
I think that is fine. LGTM. Thanks for the fix @twdsilva!

PR #27, opened 2017-01-20
TEPHRA-210 Get table specific properties from tableDescriptor and oth…

gokulavasan:
…er properties from getConfiguration method allowing users to override it to supply it

JIRA : https://issues.apache.org/jira/browse/TEPHRA-210

Todo: Port this change to other compat modules after review of this change


gokulavasan:
@poornachandra @chtyim Please take a look. Thanks!

gokulavasan:
This PR needs more work - to read the configuration for compaction lazily. Will reopen after implementing that.

PR #28, opened 2017-01-20
TEPHRA-210 Get table specific properties from tableDescriptor

gokulavasan:
And other properties from getConfiguration method, construct compactionState lazily

JIRA : https://issues.apache.org/jira/browse/TEPHRA-210

### DIFF
--- a/tephra-hbase-compat-1.1-base/src/main/java/org/apache/tephra/hbase/coprocessor/TransactionProcessor.java
+++ b/tephra-hbase-compat-1.1-base/src/main/java/org/apache/tephra/hbase/coprocessor/TransactionProcessor.java
@@ -23,6 +23,7 @@ import com.google.common.collect.Maps;
 import com.google.common.collect.Sets;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
+import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.hbase.Cell;
 import org.apache.hadoop.hbase.CellUtil;
 import org.apache.hadoop.hbase.CoprocessorEnvironment;
@@ -103,13 +104,14 @@ import javax.annotation.Nullable;
 public class TransactionProcessor extends BaseRegionObserver {
   private static final Log LOG = LogFactory.getLog(TransactionProcessor.class);
 
-  private TransactionStateCache cache;
   private final TransactionCodec txCodec;
+  private TransactionStateCache cache;
   private CompactionState compactionState;
   protected Map<byte[], Long> ttlByFamily = Maps.newTreeMap(Bytes.BYTES_COMPARATOR);
   protected boolean allowEmptyValues = TxConstants.ALLOW_EMPTY_VALUES_DEFAULT;
   protected boolean readNonTxnData = TxConstants.DEFAULT_READ_NON_TX_DATA;
-  protected long txMaxLifetimeMillis = TimeUnit.SECONDS.toMillis(TxConstants.Manager.DEFAULT_TX_MAX_LIFETIME);
+  protected Long txMaxLifetimeMillis;
+  private Boolean pruneEnable;
 
   public TransactionProcessor() {
     this.txCodec = new TransactionCodec();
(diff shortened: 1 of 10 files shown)


tephra-hbase-compat-1.1-base/src/main/java/org/apache/tephra/hbase/coprocessor/TransactionProcessor.java
```diff
@@ -140,29 +141,25 @@ public void start(CoprocessorEnvironment e) throws IOException {
         ttlByFamily.put(columnDesc.getName(), ttl);
       }
 
-      this.allowEmptyValues = env.getConfiguration().getBoolean(TxConstants.ALLOW_EMPTY_VALUES_KEY,
-                                                                TxConstants.ALLOW_EMPTY_VALUES_DEFAULT);
+      String allowEmptyValuesFromTableDesc = tableDesc.getValue(TxConstants.ALLOW_EMPTY_VALUES_KEY);
+      this.allowEmptyValues = (allowEmptyValuesFromTableDesc != null) ? Boolean.valueOf(allowEmptyValuesFromTableDesc) :
+        getConfiguration(env).getBoolean(TxConstants.ALLOW_EMPTY_VALUES_KEY, TxConstants.ALLOW_EMPTY_VALUES_DEFAULT);
```
poornachandra: Shouldn't this property also be lazily read?

tephra-hbase-compat-1.1-base/src/main/java/org/apache/tephra/hbase/coprocessor/TransactionProcessor.java
```diff
@@ -140,29 +141,25 @@ public void start(CoprocessorEnvironment e) throws IOException {
...
-      boolean pruneEnabled = env.getConfiguration().getBoolean(TxConstants.TransactionPruning.PRUNE_ENABLE,
-                                                               TxConstants.TransactionPruning.DEFAULT_PRUNE_ENABLE);
-      if (pruneEnabled) {
-        String pruneTable = env.getConfiguration().get(TxConstants.TransactionPruning.PRUNE_STATE_TABLE,
-                                                       TxConstants.TransactionPruning.DEFAULT_PRUNE_STATE_TABLE);
-        compactionState = new CompactionState(env, TableName.valueOf(pruneTable), txMaxLifetimeMillis);
-        LOG.debug("Automatic invalid list pruning is enabled. Compaction state will be recorded in table " +
-                    pruneTable);
-      }
+        TimeUnit.SECONDS.toMillis(getConfiguration(env).getInt(TxConstants.Manager.CFG_TX_MAX_LIFETIME,
```
poornachandra: Same here about lazy read

tephra-hbase-compat-1.1-base/src/main/java/org/apache/tephra/hbase/coprocessor/TransactionProcessor.java
```diff
@@ -140,29 +141,25 @@ public void start(CoprocessorEnvironment e) throws IOException {
...
+                                                               TxConstants.Manager.DEFAULT_TX_MAX_LIFETIME));
     }
   }
 
+  protected Configuration getConfiguration(CoprocessorEnvironment env) {
```
poornachandra: It would be good to add comments on the semantics of this method

tephra-hbase-compat-1.1-base/src/main/java/org/apache/tephra/hbase/coprocessor/TransactionProcessor.java
gokulavasan: I think this property is used in preGetOp. Ideally, this should be read only from the tableDesc and set. If we do it lazily, then we might get different results for a Get operation based on whether we were able to read the Configuration successfully or not. I can modify this to read it only from tableDesc similar to the other property (READ_NON_TX_DATA)

tephra-hbase-compat-1.1-base/src/main/java/org/apache/tephra/hbase/coprocessor/TransactionProcessor.java
gokulavasan: This is being lazily read again when compaction is being called. Unfortunately we have to set it in the start method since it is in ensureValidTxLifetime method and thus we need to set it in the beginning.

tephra-hbase-compat-1.1-base/src/main/java/org/apache/tephra/hbase/coprocessor/TransactionProcessor.java
gokulavasan: Sure, will add comments for this.

tephra-hbase-compat-1.1-base/src/main/java/org/apache/tephra/hbase/coprocessor/TransactionProcessor.java
poornachandra: Looks like `allowEmptyValues` was not a table level property before. Any reason why we made it into a table level property now?

> If we do it lazily, then we might get different results for a Get operation based on whether we were able to read the Configuration successfully or not.

If not then it can be initialized incorrectly during startup, and will always be wrong, right? I think it should always be read lazily.

gokulavasan: I have updated the PR to read the value from table descriptor first. If it is not available in the tableDescriptor, we try to obtain it from the configuration as we used to earlier.

gokulavasan:
@poornachandra Updated the PR by addressing comments. Please take a look again.

tephra-hbase-compat-1.1-base/src/main/java/org/apache/tephra/hbase/coprocessor/TransactionProcessor.java
```diff
@@ -303,10 +305,22 @@ public InternalScanner preCompactScannerOpen(ObserverContext<RegionCoprocessorEn
     // Get the latest tx snapshot state for the compaction
     TransactionVisibilityState snapshot = cache.getLatestState();
 
-    // Record tx state before the compaction
-    if (compactionState != null) {
+    if (pruneEnable == null) {
+      Configuration conf = getConfiguration(c.getEnvironment());
+      pruneEnable = conf.getBoolean(TxConstants.TransactionPruning.PRUNE_ENABLE,
```
poornachandra: If the configuration is not available for the first time then won't this become false for the whole run? Should we return null from `getConfiguration` in case configuration is not available? This way we can skip the run

gokulavasan: In Tephra, it will always be available right?

poornachandra: Since this code does not change between Tehpra and derived class implementation, we may need to handle the case when configuration is not available after startup. What do you think?

gokulavasan:
@poornachandra Addressed your comment by making getConfiguration method return a Configuration which could be nullable. Please look at ``ensureValidTxLifetime`` method since we now will have to throw an exception if Configuration is not available. Is that reasonable way to handle it?

tephra-hbase-compat-1.1-base/src/main/java/org/apache/tephra/hbase/coprocessor/TransactionProcessor.java
```diff
@@ -140,29 +143,37 @@ public void start(CoprocessorEnvironment e) throws IOException {
...
-      this.allowEmptyValues = env.getConfiguration().getBoolean(TxConstants.ALLOW_EMPTY_VALUES_KEY,
-                                                                TxConstants.ALLOW_EMPTY_VALUES_DEFAULT);
+      String allowEmptyValuesFromTableDesc = tableDesc.getValue(TxConstants.ALLOW_EMPTY_VALUES_KEY);
+      Configuration conf = getConfiguration(env);
+      boolean allowEmptyValuesFromConfig = (conf != null) ?
+        conf.getBoolean(TxConstants.ALLOW_EMPTY_VALUES_KEY, TxConstants.ALLOW_EMPTY_VALUES_DEFAULT) :
+        TxConstants.ALLOW_EMPTY_VALUES_DEFAULT;
+
+      // If the property is not present in the tableDescriptor, get it from the Configuration
+      this.allowEmptyValues = (allowEmptyValuesFromTableDesc != null) ?
```
poornachandra: It might help readability if we move reading `allowEmptyValues` into a private method

poornachandra:
LGTM

gokulavasan:
@poornachandra Please review the other commits. Ported changes to other compat modules.

poornachandra:
LGTM

PR #30, opened 2017-02-02
Use getRegion instead of getRegionInfo for hbase-0.98 compat module

gokulavasan:
Use the older ``getRegion()`` method instead of the ``getRegionInfo()`` method


poornachandra:
What about 0.96 module?

gokulavasan:
@poornachandra We use ``getRegion()`` in 0.96.

poornachandra:
LGTM

PR #31, opened 2017-02-07
(TEPHRA-214) Tool to debug the state and progress of Invalid List Pruning

gokulavasan:
JIRA : https://issues.apache.org/jira/browse/TEPHRA-214

Summary: Introduce a debug tool to fetch data that might help with debugging the status and progress of invalid list transaction list pruning.

Three commands are supported:
timeregion <timestamp>  ==> Will return TimeRegions at or before timestamp in millis
idleregions <numOfRegions> ==> Will return numOfRegions regions with lowest pruneUpperBounds. If -1 is provided, all pruneUpperBounds are returned.
pruneinfo <regionNameAsString> => Will return pruneUpperBound, compactionTimeStamp of the 'regionNameAsString' region.

This tool can be invoked with hbase classpath and tephra jars in the classpath.

### DIFF
--- a/tephra-hbase-compat-1.1-base/src/main/java/org/apache/tephra/hbase/txprune/DataJanitorState.java
+++ b/tephra-hbase-compat-1.1-base/src/main/java/org/apache/tephra/hbase/txprune/DataJanitorState.java
@@ -20,6 +20,8 @@
 package org.apache.tephra.hbase.txprune;
 
 import com.google.common.collect.Maps;
+import org.apache.hadoop.hbase.Cell;
+import org.apache.hadoop.hbase.CellUtil;
 import org.apache.hadoop.hbase.client.Delete;
 import org.apache.hadoop.hbase.client.Get;
 import org.apache.hadoop.hbase.client.Put;
@@ -29,8 +31,11 @@ import org.apache.hadoop.hbase.client.Scan;
 import org.apache.hadoop.hbase.client.Table;
 import org.apache.hadoop.hbase.util.Bytes;
 import org.apache.tephra.hbase.coprocessor.TransactionProcessor;
+import org.apache.tephra.txprune.RegionPruneInfo;
 
 import java.io.IOException;
+import java.util.ArrayList;
+import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.SortedSet;
@@ -102,11 +107,30 @@ public class DataJanitorState {
    * @throws IOException when not able to read the data from HBase
    */
   public long getPruneUpperBoundForRegion(byte[] regionId) throws IOException {
+    RegionPruneInfo regionPruneInfo = getPruneInfoForRegion(regionId);
+    return (regionPruneInfo == null) ? -1 : regionPruneInfo.getPruneUpperBound();
+  }
+
+  /**
+   * Get the latest {@link RegionPruneInfo} for a given region.
+   *
+   * @param regionId region id
+   * @return {@link RegionPruneInfo} for the region
+   * @throws IOException when not able to read the data from HBase
+   */
+  @Nullable
+  public RegionPruneInfo getPruneInfoForRegion(byte[] regionId) throws IOException {
     try (Table stateTable = stateTableSupplier.get()) {
       Get get = new Get(makeRegionKey(regionId));
       get.addColumn(FAMILY, PRUNE_UPPER_BOUND_COL);
-      byte[] result = stateTable.get(get).getValue(FAMILY, PRUNE_UPPER_BOUND_COL);
-      return result == null ? -1 : Bytes.toLong(result);
+      Cell cell = stateTable.get(get).getColumnLatestCell(FAMILY, PRUNE_UPPER_BOUND_COL);
+      if (cell == null) {
+        return null;
+      }
+      byte[] pruneUpperBoundBytes = CellUtil.cloneValue(cell);
+      long timestamp = cell.getTimestamp();
+      return new RegionPruneInfo(regionId, Bytes.toStringBinary(regionId),
+                                 Bytes.toLong(pruneUpperBoundBytes), timestamp);
     }
   }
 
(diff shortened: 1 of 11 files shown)


tephra-hbase-compat-1.1-base/src/main/java/org/apache/tephra/hbase/txprune/DataJanitorState.java
```diff
@@ -102,11 +107,29 @@ public void savePruneUpperBoundForRegion(byte[] regionId, long pruneUpperBound)
    * @throws IOException when not able to read the data from HBase
    */
   public long getPruneUpperBoundForRegion(byte[] regionId) throws IOException {
+    RegionPruneInfo regionPruneInfo = getPruneInfoForRegion(regionId);
+    return (regionPruneInfo == null) ? -1 : regionPruneInfo.getCompactionTimestamp();
```
poornachandra: This has to return the `regionPruneInfo.getPruneUpperBound()`, right?

tephra-hbase-compat-1.1-base/src/main/java/org/apache/tephra/hbase/txprune/DataJanitorState.java
```diff
@@ -102,11 +107,29 @@ public void savePruneUpperBoundForRegion(byte[] regionId, long pruneUpperBound)
...
+  }
+
+  /**
+   * Get the latest {@link RegionPruneInfo} for a given region.
+   *
+   * @param regionId region id
+   * @return {@link RegionPruneInfo} for the region
+   * @throws IOException when not able to read the data from HBase
+   */
+  public RegionPruneInfo getPruneInfoForRegion(byte[] regionId) throws IOException {
```
poornachandra: This method needs to be annotated as `@Nullable`

tephra-hbase-compat-1.1-base/src/main/java/org/apache/tephra/hbase/txprune/DataJanitorState.java
```diff
@@ -120,6 +143,22 @@ public long getPruneUpperBoundForRegion(byte[] regionId) throws IOException {
    */
   public Map<byte[], Long> getPruneUpperBoundForRegions(SortedSet<byte[]> regions) throws IOException {
     Map<byte[], Long> resultMap = new TreeMap<>(Bytes.BYTES_COMPARATOR);
+    List<RegionPruneInfo> regionPruneInfos = getPruneInfoForRegions(regions);
+    for (RegionPruneInfo regionPruneInfo : regionPruneInfos) {
+      resultMap.put(regionPruneInfo.getRegionName(), regionPruneInfo.getCompactionTimestamp());
```
poornachandra: Same here, should be `regionPruneInfo.getPruneUpperBound()` instead of `regionPruneInfo.getCompactionTimestamp()`

tephra-hbase-compat-1.1-base/src/main/java/org/apache/tephra/hbase/txprune/DataJanitorState.java
```diff
@@ -129,17 +168,19 @@ public long getPruneUpperBoundForRegion(byte[] regionId) throws IOException {
...
-          if (regions.contains(region)) {
-            byte[] timeBytes = next.getValue(FAMILY, PRUNE_UPPER_BOUND_COL);
-            if (timeBytes != null) {
-              long pruneUpperBoundRegion = Bytes.toLong(timeBytes);
-              resultMap.put(region, pruneUpperBoundRegion);
+          if (regions == null || regions.contains(region)) {
+            Cell cell = next.getColumnLatestCell(FAMILY, PRUNE_UPPER_BOUND_COL);
+            if (cell != null) {
+              byte[] pruneUpperBoundBytes = CellUtil.cloneValue(cell);
+              long timestamp = cell.getTimestamp();
```
poornachandra: This is not compaction timestamp, this is when the prune upper bound was written to HBase. Although this can be an approximate value when the state table is available. Is the compaction timestamp really necessary for debugging?

poornachandra: Also it may be okay to treat this as the compaction timestamp for now.

tephra-hbase-compat-1.1-base/src/main/java/org/apache/tephra/hbase/txprune/InvalidListPruningDebug.java
```diff
@@ -0,0 +1,191 @@
...
+    }
+    return timeRegions.toString();
+  }
+
+  private void printUsage(PrintWriter pw) {
+    pw.println("Usage : org.apache.tephra.hbase.txprune.InvalidListPruning <command> <parameter>");
+    pw.println("Available commands, corresponding parameters are:");
+    pw.println("****************************************************");
+    pw.println("timeregion ts");
+    pw.println("Desc: Prints out a time region at time 'ts' (in milliseconds) or the latest before time 'ts'.");
```
poornachandra: `Prints out the transactional regions present in HBase at time...`

poornachandra:
@gokulavasan I took a pass over this, I have a few comments.

tephra-hbase-compat-1.1-base/src/main/java/org/apache/tephra/hbase/txprune/DataJanitorState.java
gokulavasan: @poornachandra It might be useful information to have right? I can rename it to something else if you think compaction timestamp is misleading.

tephra-hbase-compat-1.1-base/src/main/java/org/apache/tephra/hbase/txprune/HBaseTransactionPruningPlugin.java
```diff
@@ -147,6 +147,7 @@ public long fetchPruneUpperBound(long time, long inactiveTransactionBound) throw
 
     // Get all the current transactional regions
     SortedSet<byte[]> transactionalRegions = getTransactionalRegions();
+    // TODO: Ignore empty
```
gokulavasan: That TODO was meant for ignoring empty regions from the current transactional regions. I have removed that misleading TODO as part of the last commit.

tephra-hbase-compat-1.1-base/src/main/java/org/apache/tephra/hbase/txprune/DataJanitorState.java
poornachandra: What would be a good name for this?

gokulavasan: PruneRecordTime ?

poornachandra:
LGTM

gokulavasan:
@poornachandra Ported changes to other modules. Please take a look. Thanks!

poornachandra:
LGTM

PR #32, opened 2017-02-07
(TEPHRA-215) (TEPHRA-218) Use single thread across all regions in a r…

gokulavasan:
…egion server to persist Prune Upper Bound info. Also don't refresh cache during startup of TransactionStateCache to avoid the possibility of Service stopping if tx.snapshot dir is not found during startup

JIRA : 
https://issues.apache.org/jira/browse/TEPHRA-215
https://issues.apache.org/jira/browse/TEPHRA-218

### DIFF
--- a/tephra-hbase-compat-1.1-base/src/main/java/org/apache/tephra/hbase/txprune/PruneUpperBoundWriter.java
+++ b/tephra-hbase-compat-1.1-base/src/main/java/org/apache/tephra/hbase/txprune/PruneUpperBoundWriter.java
@@ -18,57 +18,62 @@
 
 package org.apache.tephra.hbase.txprune;
 
+import com.google.common.util.concurrent.AbstractIdleService;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.apache.hadoop.hbase.TableName;
-import org.apache.tephra.hbase.coprocessor.TransactionProcessor;
+import org.apache.hadoop.hbase.util.Bytes;
 
 import java.io.IOException;
+import java.util.Map;
+import java.util.concurrent.ConcurrentSkipListMap;
 import java.util.concurrent.TimeUnit;
-import java.util.concurrent.atomic.AtomicBoolean;
-import java.util.concurrent.atomic.AtomicLong;
 
 /**
- * Thread that will write the the prune upper bound
+ * Thread that will write the the prune upper bound. An instance of this class should be obtained only
+ * through {@link PruneUpperBoundWriterSupplier} which will also handle the lifecycle of this instance.
  */
-public class PruneUpperBoundWriter {
-  private static final Log LOG = LogFactory.getLog(TransactionProcessor.class);
+public class PruneUpperBoundWriter extends AbstractIdleService {
+  private static final Log LOG = LogFactory.getLog(PruneUpperBoundWriter.class);
 
-  private final TableName pruneStateTable;
+  private final TableName tableName;
   private final DataJanitorState dataJanitorState;
-  private final byte[] regionName;
-  private final String regionNameAsString;
   private final long pruneFlushInterval;
-  private final AtomicLong pruneUpperBound;
-  private final AtomicBoolean shouldFlush;
+  private final ConcurrentSkipListMap<byte[], Long> pruneEntries;
 
-  private Thread flushThread;
+  private volatile Thread flushThread;
+
   private long lastChecked;
 
-  public PruneUpperBoundWriter(DataJanitorState dataJanitorState, TableName pruneStateTable, String regionNameAsString,
-                               byte[] regionName, long pruneFlushInterval) {
-    this.pruneStateTable = pruneStateTable;
+  public PruneUpperBoundWriter(TableName tableName, DataJanitorState dataJanitorState, long pruneFlushInterval) {
+    this.tableName = tableName;
     this.dataJanitorState = dataJanitorState;
-    this.regionName = regionName;
-    this.regionNameAsString = regionNameAsString;
     this.pruneFlushInterval = pruneFlushInterval;
-    this.pruneUpperBound = new AtomicLong();
-    this.shouldFlush = new AtomicBoolean(false);
-    startFlushThread();
+    this.pruneEntries = new ConcurrentSkipListMap<>(Bytes.BYTES_COMPARATOR);
+  }
+
+  public void persistPruneEntry(byte[] regionName, long pruneUpperBound) {
+    // The number of entries in this map is bound by the number of regions in this region server and thus it will not
+    // grow indefinitely
+    pruneEntries.put(regionName, pruneUpperBound);
   }
 
   public boolean isAlive() {
-    return flushThread.isAlive();
+    return flushThread != null && flushThread.isAlive();
   }
 
-  public void persistPruneEntry(long pruneUpperBound) {
-    this.pruneUpperBound.set(pruneUpperBound);
-    this.shouldFlush.set(true);
+  @Override
+  protected void startUp() throws Exception {
+    LOG.info("Starting PruneUpperBoundWriter Thread.");
+    startFlushThread();
   }
 
-  public void stop() {
+  @Override
+  protected void shutDown() throws Exception {
+    LOG.info("Stopping PruneUpperBoundWriter Thread.");
     if (flushThread != null) {
       flushThread.interrupt();
+      flushThread.join(TimeUnit.SECONDS.toMillis(1));
     }
   }
 
@@ -76,19 +81,21 @@ public class PruneUpperBoundWriter {
     flushThread = new Thread("tephra-prune-upper-bound-writer") {
       @Override
       public void run() {
-        while (!isInterrupted()) {
+        while ((!isInterrupted()) && isRunning()) {
           long now = System.currentTimeMillis();
           if (now > (lastChecked + pruneFlushInterval)) {
-            if (shouldFlush.compareAndSet(true, false)) {
-              // should flush data
-              try {
-                dataJanitorState.savePruneUpperBoundForRegion(regionName, pruneUpperBound.get());
-              } catch (IOException ex) {
-                LOG.warn("Cannot record prune upper bound for region " + regionNameAsString + " in the table " +
-                           pruneStateTable.getNameWithNamespaceInclAsString() + " after compacting region.", ex);
-                // Retry again
-                shouldFlush.set(true);
+            // should flush data
+            try {
+              while (pruneEntries.firstEntry() != null) {
+                Map.Entry<byte[], Long> firstEntry = pruneEntries.firstEntry();
+                dataJanitorState.savePruneUpperBoundForRegion(firstEntry.getKey(), firstEntry.getValue());
+                // We can now remove the entry only if the key and value match with what we wrote since it is
+                // possible that a new pruneUpperBound for the same key has been added
+                pruneEntries.remove(firstEntry.getKey(), firstEntry.getValue());
               }
+            } catch (IOException ex) {
+              LOG.warn("Cannot record prune upper bound for a region to table " +
+                         tableName.getNameWithNamespaceInclAsString(), ex);
             }
             lastChecked = now;
           }
--- a/tephra-hbase-compat-1.1-base/src/main/java/org/apache/tephra/hbase/txprune/PruneUpperBoundWriterSupplier.java
+++ b/tephra-hbase-compat-1.1-base/src/main/java/org/apache/tephra/hbase/txprune/PruneUpperBoundWriterSupplier.java
@@ -1,0 +1,89 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.tephra.hbase.txprune;
+
+
+import com.google.common.base.Supplier;
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import org.apache.hadoop.hbase.TableName;
+
+/**
+ * Supplies instances of {@link PruneUpperBoundWriter} implementations.
+ */
+public class PruneUpperBoundWriterSupplier implements Supplier<PruneUpperBoundWriter> {
+  private static final Log LOG = LogFactory.getLog(PruneUpperBoundWriterSupplier.class);
+
+  private static volatile PruneUpperBoundWriter instance;
+  private static volatile int refCount = 0;
+  private static final Object lock = new Object();
+
+  private final TableName tableName;
+  private final DataJanitorState dataJanitorState;
+  private final long pruneFlushInterval;
+
+  public PruneUpperBoundWriterSupplier(TableName tableName, DataJanitorState dataJanitorState,
+                                       long pruneFlushInterval) {
+    this.tableName = tableName;
+    this.dataJanitorState = dataJanitorState;
+    this.pruneFlushInterval = pruneFlushInterval;
+  }
+
+  @Override
+  public PruneUpperBoundWriter get() {
+    synchronized (lock) {
+      if (instance == null) {
+        instance = new PruneUpperBoundWriter(tableName, dataJanitorState, pruneFlushInterval);
+        instance.startAndWait();
+      }
+      refCount++;
+      if (LOG.isDebugEnabled()) {
+        LOG.debug("Incrementing Reference Count for PruneUpperBoundWriter " + refCount);
+      }
+      return instance;
+    }
+  }
+
+  public void release() {
+    synchronized (lock) {
+      refCount--;
+      if (LOG.isDebugEnabled()) {
+        LOG.debug("Decrementing Reference Count for PruneUpperBoundWriter " + refCount);
+      }
+
+      if (refCount == 0) {
+        try {
+          instance.stopAndWait();
+        } catch (Exception ex) {
+          LOG.warn("Exception while trying to shutdown PruneUpperBoundWriter thread. ", ex);
+        } finally {
+          // If the thread is still alive (might happen if the thread was blocked on HBase PUT call), interrupt it again
+          if (instance.isAlive()) {
+            try {
+              instance.shutDown();
+            } catch (Exception e) {
+              LOG.warn("Exception while trying to shutdown PruneUpperBoundWriter thread. ", e);
+            }
+          }
+          instance = null;
+        }
+      }
+    }
+  }
+}
--- a/tephra-hbase-compat-1.1-base/src/test/java/org/apache/tephra/hbase/txprune/PruneUpperBoundWriterSupplierTest.java
+++ b/tephra-hbase-compat-1.1-base/src/test/java/org/apache/tephra/hbase/txprune/PruneUpperBoundWriterSupplierTest.java
@@ -1,0 +1,122 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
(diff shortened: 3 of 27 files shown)


tephra-hbase-compat-1.1-base/src/main/java/org/apache/tephra/hbase/txprune/PruneUpperBoundWriterSupplier.java
```diff
@@ -0,0 +1,52 @@
...
+
+/**
+ * Supplies instances of {@link PruneUpperBoundWriter} implementations.
+ */
+public class PruneUpperBoundWriterSupplier implements Supplier<PruneUpperBoundWriter> {
+
+  private final DataJanitorState dataJanitorState;
+  private final long pruneFlushInterval;
+
+  protected static volatile PruneUpperBoundWriter instance;
```
poornachandra: I don't think this needs to be protected

tephra-hbase-compat-1.1-base/src/main/java/org/apache/tephra/hbase/txprune/PruneUpperBoundWriterSupplier.java
```diff
@@ -0,0 +1,52 @@
...
+  protected static Object lock = new Object();
```
poornachandra: Same here about protected. Also it is better to make the `lock` object final.

tephra-hbase-compat-1.1-base/src/main/java/org/apache/tephra/hbase/txprune/PruneUpperBoundWriter.java
```diff
@@ -18,55 +18,48 @@
...
-    this.pruneStateTable = pruneStateTable;
+  public PruneUpperBoundWriter(DataJanitorState dataJanitorState, long pruneFlushInterval) {
     this.dataJanitorState = dataJanitorState;
-    this.regionName = regionName;
-    this.regionNameAsString = regionNameAsString;
     this.pruneFlushInterval = pruneFlushInterval;
-    this.pruneUpperBound = new AtomicLong();
-    this.shouldFlush = new AtomicBoolean(false);
+    this.pruneEntries = new ConcurrentLinkedQueue<>();
     startFlushThread();
```
poornachandra: Since the flush thread is started in the `startUp` method, we don't need to start it in the constructor.

tephra-hbase-compat-1.1-base/src/main/java/org/apache/tephra/hbase/txprune/PruneUpperBoundWriter.java
```diff
@@ -18,55 +18,48 @@
...
   }
 
-  public boolean isAlive() {
-    return flushThread.isAlive();
+  public void persistPruneEntry(byte[] regionName, long pruneUpperBound) {
+    pruneEntries.add(new PruneInfo(regionName, pruneUpperBound));
```
poornachandra: I think it would be good to limit the number of entries in the `pruneEntries` queue. If we are over the limit then we can drop earlier elements.

tephra-hbase-compat-1.1-base/src/main/java/org/apache/tephra/hbase/txprune/PruneUpperBoundWriter.java
```diff
@@ -79,15 +72,17 @@ public void run() {
         while (!isInterrupted()) {
           long now = System.currentTimeMillis();
           if (now > (lastChecked + pruneFlushInterval)) {
-            if (shouldFlush.compareAndSet(true, false)) {
+            if (!pruneEntries.isEmpty()) {
               // should flush data
               try {
-                dataJanitorState.savePruneUpperBoundForRegion(regionName, pruneUpperBound.get());
+                Iterator<PruneInfo> iterator = pruneEntries.iterator();
```
poornachandra: Since we are doing concurrent operations here it would be good to use `peek()` and `poll()`, rather than a iterator.

tephra-hbase-compat-1.1-base/src/main/java/org/apache/tephra/hbase/txprune/PruneUpperBoundWriterSupplier.java
```diff
@@ -0,0 +1,52 @@
...
+    this.pruneFlushInterval = pruneFlushInterval;
+  }
+
+  @Override
+  public PruneUpperBoundWriter get() {
+    if (instance == null) {
+      synchronized (lock) {
+        if (instance == null) {
+          instance = new PruneUpperBoundWriter(dataJanitorState, pruneFlushInterval);
+          instance.start();
```
poornachandra: We have the same problem as `TransactionStateCache` here too - we don't have a way to stop this thread after all instances of `TransactionProcessor` co-processors have been shutdown (https://issues.apache.org/jira/browse/TEPHRA-152)

gokulavasan: @poornachandra Do you have any thoughts on how to address this?

tephra-hbase-compat-1.1-base/src/main/java/org/apache/tephra/hbase/txprune/PruneUpperBoundWriter.java
gokulavasan: @poornachandra Should that max size be configurable?

gokulavasan:
@poornachandra Can you please take a look again. Thanks!

gokulavasan:
@poornachandra Please review when you get a chance. Thanks!

tephra-hbase-compat-1.1-base/src/main/java/org/apache/tephra/hbase/txprune/PruneUpperBoundWriter.java
```diff
@@ -18,56 +18,56 @@
...
+  private static final Log LOG = LogFactory.getLog(PruneUpperBoundWriter.class);
 
-  private final TableName pruneStateTable;
   private final DataJanitorState dataJanitorState;
-  private final byte[] regionName;
-  private final String regionNameAsString;
   private final long pruneFlushInterval;
-  private final AtomicLong pruneUpperBound;
-  private final AtomicBoolean shouldFlush;
+  private final ConcurrentSkipListMap<ByteBuffer, Long> pruneEntries;
```
poornachandra: It would be simpler to use `ConcurrentSkipListMap<byte[], Long>`

tephra-hbase-compat-1.1-base/src/main/java/org/apache/tephra/hbase/txprune/PruneUpperBoundWriter.java
```diff
@@ -76,18 +76,21 @@ private void startFlushThread() {
...
+                  // We can now remove the entry only if the key and value match with what we wrote since it is
+                  // possible that a new pruneUpperBound for the same key has been added
+                  pruneEntries.remove(firstEntry.getKey(), firstEntry.getValue());
+                }
               } catch (IOException ex) {
-                LOG.warn("Cannot record prune upper bound for region " + regionNameAsString + " in the table " +
-                           pruneStateTable.getNameWithNamespaceInclAsString() + " after compacting region.", ex);
-                // Retry again
-                shouldFlush.set(true);
+                LOG.warn("Cannot record prune upper bound for a region.", ex);
```
poornachandra: It would be good to add prune state table name to the log message, so that it can help in figuring out mis-configurations if any

tephra-hbase-compat-1.1-base/src/main/java/org/apache/tephra/hbase/txprune/PruneUpperBoundWriterSupplier.java
```diff
@@ -0,0 +1,78 @@
...
+  private final long pruneFlushInterval;
+
+  public PruneUpperBoundWriterSupplier(DataJanitorState dataJanitorState, long pruneFlushInterval) {
+    this.dataJanitorState = dataJanitorState;
+    this.pruneFlushInterval = pruneFlushInterval;
+  }
+
+  @Override
+  public PruneUpperBoundWriter get() {
+    synchronized (lock) {
```
poornachandra: We can make an `instance == null` check outside the lock. In case instance is not null just increment refCount and return.

tephra-hbase-compat-1.1-base/src/main/java/org/apache/tephra/hbase/txprune/PruneUpperBoundWriterSupplier.java
```diff
@@ -0,0 +1,78 @@
...
+        instance.startAndWait();
+      }
+      refCount++;
+      return instance;
+    }
+  }
+
+  public void release() {
+    synchronized (lock) {
+      refCount--;
```
poornachandra: Same here, decrement refCount outside lock and return if refCount is greater than zero. Only lock when stopping.

tephra-hbase-compat-1.1-base/src/test/java/org/apache/tephra/hbase/txprune/InvalidListPruneTest.java
```diff
@@ -360,8 +404,18 @@ public TransactionStateCache get() {
...
+      synchronized (compactionLock) {
+        super.postCompact(e, store, resultFile, request);
+        lastMajorCompactionTime.set(System.currentTimeMillis());
+      }
+    }
+
+    @Override
+    public void stop(CoprocessorEnvironment e) throws IOException {
+      synchronized (stopLock) {
+        System.out.println("*********** Stopping " + this.getClass().getName());
```
poornachandra: Remove the debug println

tephra-hbase-compat-1.1-base/src/test/java/org/apache/tephra/hbase/txprune/InvalidListPruneTest.java
```diff
@@ -334,6 +337,45 @@ public Table get() throws IOException {
...
 
+  @Test(timeout = 60000L)
+  public void testClusterShutdown() throws Exception {
+    java.util.concurrent.ExecutorService executorService = Executors.newSingleThreadExecutor();
+    try {
+      // Create a new transaction snapshot
+      InMemoryTransactionStateCache.setTransactionSnapshot(
+        new TransactionSnapshot(100, 100, 100, ImmutableSet.of(50L),
+                                ImmutableSortedMap.<Long, TransactionManager.InProgressTx>of()));
+      // Run major compaction
```
poornachandra: The comment can be changed to - `Try to create a situation where HBase shuts down when the async prune writer thread is trying to write to the prune table on a compaction. This will surface any deadlocks during HBase shutdown.`

tephra-hbase-compat-1.1-base/src/test/java/org/apache/tephra/hbase/txprune/PruneUpperBoundWriterSupplierTest.java
```diff
@@ -0,0 +1,119 @@
...
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
+
+/**
+ * Tests to verify the behavior of {@link PruneUpperBoundWriterSupplier}.
+ */
+public class PruneUpperBoundWriterSupplierTest extends AbstractHBaseTableTest {
```
poornachandra: Can you add comments on what this test is trying to do? I'm not sure I follow the test.

Also, do we need to start an HBase instance for this test? We just need to make sure that prune writer gets stopped when reference count reaches zero, right? In that case can we pass in DataJanitorState as null to PruneUpperBoundWriterSupplier, and not call persistPruneEntry?

tephra-hbase-compat-1.1-base/src/main/java/org/apache/tephra/hbase/txprune/PruneUpperBoundWriter.java
```diff
@@ -18,56 +18,56 @@
...
-    this.shouldFlush.set(true);
+  @Override
+  protected void startUp() throws Exception {
+    startFlushThread();
   }
 
-  public void stop() {
+  @Override
+  protected void shutDown() throws Exception {
     if (flushThread != null) {
```
poornachandra: Also would be good to add a log statement saying "stopping PruneUpperBoundWriter" and "stopped PruneUpperBoundWriter"

tephra-hbase-compat-1.1-base/src/main/java/org/apache/tephra/hbase/txprune/PruneUpperBoundWriter.java
gokulavasan: We need the key to be Comparable. Hence we need to use ByteBuffer.

tephra-hbase-compat-1.1-base/src/main/java/org/apache/tephra/hbase/txprune/PruneUpperBoundWriter.java
```diff
@@ -18,56 +18,56 @@
...
+      stopped = true;
       flushThread.interrupt();
     }
```
gokulavasan: What would be a good timeout?

gokulavasan:
@poornachandra Addressed comments. Please take a look when you get a chance.

tephra-hbase-compat-1.1-base/src/main/java/org/apache/tephra/hbase/txprune/PruneUpperBoundWriter.java
gokulavasan: Added log statements.

tephra-hbase-compat-1.1-base/src/main/java/org/apache/tephra/hbase/txprune/PruneUpperBoundWriterSupplier.java
```diff
@@ -0,0 +1,84 @@
...
+
+  @Override
+  public PruneUpperBoundWriter get() {
+    synchronized (lock) {
+      if (instance == null) {
+        instance = new PruneUpperBoundWriter(tableName, dataJanitorState, pruneFlushInterval);
+        instance.startAndWait();
+      }
+      refCount++;
+      LOG.debug("Incrementing Reference Count for PruneUpperBoundWriter " + refCount);
```
poornachandra: Would be good to enclose this log statement with `LOG.isDebugEnabled()`

tephra-hbase-compat-1.1-base/src/main/java/org/apache/tephra/hbase/txprune/PruneUpperBoundWriterSupplier.java
```diff
@@ -0,0 +1,84 @@
...
+      return instance;
+    }
+  }
+
+  public void release() {
+    synchronized (lock) {
+      refCount--;
+      LOG.debug("Decrementing Reference Count for PruneUpperBoundWriter " + refCount);
```
poornachandra: Same here

tephra-hbase-compat-1.1-base/src/test/java/org/apache/tephra/hbase/txprune/InvalidListPruneTest.java
```diff
@@ -334,6 +337,46 @@ public Table get() throws IOException {
     }
   }
 
+  @Test(timeout = 60000L)
+  public void testClusterShutdown() throws Exception {
```
poornachandra: Is this test required now?

tephra-hbase-compat-1.1-base/src/test/java/org/apache/tephra/hbase/txprune/PruneUpperBoundWriterSupplierTest.java
```diff
@@ -0,0 +1,105 @@
...
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicInteger;
+
+/**
+ * Tests to verify the behavior of {@link PruneUpperBoundWriterSupplier}.
+ */
+public class PruneUpperBoundWriterSupplierTest {
+  private static final int NUM_OPS = 10;
```
poornachandra: Since this is a fast operation, let's make this a bigger number. 1000 maybe?

tephra-hbase-compat-1.1-base/src/test/java/org/apache/tephra/hbase/txprune/PruneUpperBoundWriterSupplierTest.java
```diff
@@ -0,0 +1,105 @@
...
+    final PruneUpperBoundWriterSupplier supplier = new PruneUpperBoundWriterSupplier(null, null, 10L);
+    final PruneUpperBoundWriter writer = supplier.get();
+    final AtomicInteger numOps = new AtomicInteger(NUM_OPS);
+    final Random random = new Random(System.currentTimeMillis());
+
+    // Start threads that will 'get' PruneUpperBoundWriters
+    ExecutorService executor = Executors.newFixedThreadPool(3);
+    List<Future> futureList = new ArrayList<>();
+    for (int i = 0; i < 3; i++) {
+      futureList.add(executor.submit(new Callable<Void>() {
```
poornachandra: `Callable` can be replaced with `Runnable` so that the return statement is not required.

tephra-hbase-compat-1.1-base/src/test/java/org/apache/tephra/hbase/txprune/PruneUpperBoundWriterSupplierTest.java
```diff
@@ -0,0 +1,105 @@
...
+    }
+    executor.shutdown();
+    executor.awaitTermination(2, TimeUnit.SECONDS);
+
+    futureList.clear();
+    numOps.set(NUM_OPS);
+    // Start thread that release PruneUpperBoundWriters
+    executor = Executors.newFixedThreadPool(3);
+    for (int i = 0; i < 3; i++) {
+      futureList.add(executor.submit(new Callable<Void>() {
```
poornachandra: Same here about using `Runnable`

tephra-hbase-compat-1.1-base/src/test/java/org/apache/tephra/hbase/txprune/PruneUpperBoundWriterSupplierTest.java
```diff
@@ -0,0 +1,105 @@
...
+  public void testSupplier() throws Exception {
+    final PruneUpperBoundWriterSupplier supplier = new PruneUpperBoundWriterSupplier(null, null, 10L);
+    final PruneUpperBoundWriter writer = supplier.get();
+    final AtomicInteger numOps = new AtomicInteger(NUM_OPS);
+    final Random random = new Random(System.currentTimeMillis());
+
+    // Start threads that will 'get' PruneUpperBoundWriters
+    ExecutorService executor = Executors.newFixedThreadPool(3);
+    List<Future> futureList = new ArrayList<>();
+    for (int i = 0; i < 3; i++) {
```
poornachandra: It would be good to have more threads too

tephra-hbase-compat-1.1-base/src/test/java/org/apache/tephra/hbase/txprune/PruneUpperBoundWriterSupplierTest.java
```diff
@@ -0,0 +1,105 @@
...
+    // Start thread that release PruneUpperBoundWriters
+    executor = Executors.newFixedThreadPool(3);
+    for (int i = 0; i < 3; i++) {
+      futureList.add(executor.submit(new Callable<Void>() {
+
+        @Override
+        public Void call() throws Exception {
+          // We need to release all NUM_OPS 'gets' that were executed to trigger shutdown of the single instance of
+          // PruneUpperBoundWriter
+          while (numOps.decrementAndGet() >= 0) {
```
poornachandra: Can we change this to leave one instance out, so that we can assert that that instance is still running after the n-1 instances were released back. We can later release it separately and then assert that it is not running.

gokulavasan:
@poornachandra Addressed comments. Please take a look when you get a chance.

tephra-hbase-compat-1.1-base/src/test/java/org/apache/tephra/hbase/txprune/PruneUpperBoundWriterSupplierTest.java
```diff
@@ -74,30 +81,35 @@ public Void call() throws Exception {
...
-          return null;
         }
       }));
     }
 
     for (Future future : futureList) {
       future.get(1, TimeUnit.SECONDS);
     }
+    // Since we got one instance in the beginning, we need to release it
+    supplier.release();
```
poornachandra: It would be good to assert that the writer is running and alive before we do the final release

gokulavasan:
@poornachandra Ported changes to other compat modules. Please review when you get a chance.

poornachandra:
LGTM

PR #34, opened 2017-02-10
TEPHRA-216 Handle empty transactional regions during inactive list pruning

poornachandra:
JIRA - https://issues.apache.org/jira/browse/TEPHRA-216

Approach:
1. Transaction Processor records a region as empty (at current time) in the post flush handler if the memstore size for the region is zero and the region does not have any store files.
2. The regions that are recorded as empty after inactiveTransactionBoundTime will not have invalid data for transactions started before or on inactiveTransactionBoundTime. Hence we can consider the prune upper bound for these empty regions as the inactiveTransactionBound.

TODO: Clean-up empty regions in pruneComplete()

Note: The first two commits are re-factoring, and would be simple to review them separately.

### DIFF
--- a/tephra-hbase-compat-0.96/src/test/java/org/apache/tephra/hbase/txprune/InvalidListPruneTest.java
+++ b/tephra-hbase-compat-0.96/src/test/java/org/apache/tephra/hbase/txprune/InvalidListPruneTest.java
@@ -74,8 +74,9 @@ public class InvalidListPruneTest extends AbstractHBaseTableTest {
 
   private static TableName txDataTable1;
   private static TableName pruneStateTable;
+  private static DataJanitorState dataJanitorState;
 
-  private HConnection connection;
+  private static HConnection connection;
 
   // Override AbstractHBaseTableTest.startMiniCluster to setup configuration
   @BeforeClass
@@ -109,17 +110,25 @@ public class InvalidListPruneTest extends AbstractHBaseTableTest {
 
     pruneStateTable = TableName.valueOf(conf.get(TxConstants.TransactionPruning.PRUNE_STATE_TABLE,
                                                  TxConstants.TransactionPruning.DEFAULT_PRUNE_STATE_TABLE));
+    connection = HConnectionManager.createConnection(conf);
+    dataJanitorState =
+      new DataJanitorState(new DataJanitorState.TableSupplier() {
+        @Override
+        public HTableInterface get() throws IOException {
+          return connection.getTable(pruneStateTable);
+        }
+      });
   }
 
   @AfterClass
   public static void shutdownAfterClass() throws Exception {
+    connection.close();
     hBaseAdmin.disableTable(txDataTable1);
     hBaseAdmin.deleteTable(txDataTable1);
   }
 
   @Before
   public void beforeTest() throws Exception {
-    connection = HConnectionManager.createConnection(conf);
     createPruneStateTable();
     InMemoryTransactionStateCache.setTransactionSnapshot(null);
   }
@@ -133,8 +142,12 @@ public class InvalidListPruneTest extends AbstractHBaseTableTest {
 
   @After
   public void afterTest() throws Exception {
+    // Disable the data table so that prune writer thread gets stopped,
+    // this makes sure that any cached value will not interfere with next test
+    hBaseAdmin.disableTable(txDataTable1);
     deletePruneStateTable();
-    connection.close();
+    // Enabling the table enables the prune writer thread again
+    hBaseAdmin.enableTable(txDataTable1);
   }
 
   private void deletePruneStateTable() throws Exception {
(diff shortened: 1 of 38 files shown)


tephra-hbase-compat-1.1-base/src/main/java/org/apache/tephra/hbase/txprune/HBaseTransactionPruningPlugin.java
```diff
@@ -288,23 +288,27 @@ private long computePruneUpperBound(TimeRegions timeRegions) throws IOException
       SortedSet<byte[]> transactionalRegions = timeRegions.getRegions();
       long time = timeRegions.getTime();
 
+      long inactiveTransactionBound = dataJanitorState.getInactiveTransactionBoundForTime(time);
+      LOG.debug("Got inactive transaction bound {}", inactiveTransactionBound);
+      // If inactiveTransactionBound is not recorded then that means the data is not complete for these regions
+      if (inactiveTransactionBound == -1) {
+        if (LOG.isDebugEnabled()) {
```
anew: is this if necessary?

tephra-hbase-compat-1.1-base/src/main/java/org/apache/tephra/hbase/txprune/DataJanitorState.java
```diff
@@ -366,6 +371,56 @@ public void deleteInactiveTransactionBoundsOnOrBeforeTime(long time) throws IOEx
...
+   * Save the given region as empty as of the given time.
+   *
+   * @param time time in milliseconds
+   * @param regionId region id
+   */
+  public void saveEmptyRegionForTime(long time, byte[] regionId) throws IOException {
+    byte[] timeBytes = Bytes.toBytes(time);
+    try (Table stateTable = stateTableSupplier.get()) {
+      Put put = new Put(makeEmptyRegionTimeKey(timeBytes, regionId));
+      put.addColumn(FAMILY, EMPTY_REGION_TIME_COL, EMPTY_BYTE_ARRAY);
```
anew: maybe use a value that is visible to the human eye? Just thinking if you debug and scan the table, a value of "X" is easier to spot than an empty value.

tephra-hbase-compat-1.1-base/src/main/java/org/apache/tephra/hbase/txprune/HBaseTransactionPruningPlugin.java
```diff
@@ -322,6 +333,28 @@ private long computePruneUpperBound(TimeRegions timeRegions) throws IOException
...
+    LOG.debug("Got empty transactional regions for inactive transaction bound time {}: {}",
+              inactiveTransactionBoundTime, Iterables.transform(emptyRegions, TimeRegions.BYTE_ARR_TO_STRING_FN));
+
+    // The regions that are recorded as empty after inactiveTransactionBoundTime will not have invalid data
+    // for transactions started before or on inactiveTransactionBoundTime. Hence we can consider the prune upper bound
+    // for these empty regions as inactiveTransactionBound
+    Map<byte[], Long> pubWithEmptyRegions = new TreeMap<>(Bytes.BYTES_COMPARATOR);
+    pubWithEmptyRegions.putAll(pruneUpperBoundRegions);
+    for (byte[] emptyRegion : emptyRegions) {
+      if (!pruneUpperBoundRegions.containsKey(emptyRegion)) {
```
gokulavasan: Shouldn't this condition be reversed? If the region is empty, set the upper bound for it to be ``inactiveTransactionBound`` .

tephra-hbase-compat-1.1-base/src/main/java/org/apache/tephra/hbase/txprune/HBaseTransactionPruningPlugin.java
```diff
@@ -322,6 +333,28 @@ private long computePruneUpperBound(TimeRegions timeRegions) throws IOException
...
+        pubWithEmptyRegions.put(emptyRegion, inactiveTransactionBound);
```
gokulavasan: Shouldn't this condition be reversed?

tephra-hbase-compat-1.1-base/src/main/java/org/apache/tephra/hbase/txprune/HBaseTransactionPruningPlugin.java
poornachandra: You are right, it is not necessary. I'll remove it

poornachandra: Ah - now I remember. I added it avoid the string concatenation due to line break.

tephra-hbase-compat-1.1-base/src/main/java/org/apache/tephra/hbase/txprune/HBaseTransactionPruningPlugin.java
poornachandra: If an empty region does not contain a prune upper bound value then we put `inactiveTransactionBound` as its prune upper bound. How do you suggest this be changed?

tephra-hbase-compat-1.1-base/src/main/java/org/apache/tephra/hbase/txprune/DataJanitorState.java
poornachandra: Sure - will replace empty byte with '1'

tephra-hbase-compat-1.1-base/src/main/java/org/apache/tephra/hbase/txprune/HBaseTransactionPruningPlugin.java
gokulavasan: Ah I see sorry misunderstood the comment.

poornachandra:
@anew @gokulavasan In addition to addressing comments, I have added two more commits, please take a look

anew:
LGTM

poornachandra:
Thanks @anew. I will squash the commits and then port changes to other compat modules.

poornachandra:
@gokulavasan I have ported the changes to other compat modules, please take a look.

tephra-hbase-compat-0.96/src/test/java/org/apache/tephra/hbase/txprune/InvalidListPruneTest.java
```diff
@@ -341,6 +310,87 @@ public HTableInterface get() throws IOException {
...
+      TimeUnit.SECONDS.sleep(2);
+      pruneUpperBound2 = transactionPruningPlugin.fetchPruneUpperBound(now2, inactiveTxTimeNow2);
+      Assert.assertEquals(expectedPruneUpperBound2, pruneUpperBound2);
+      transactionPruningPlugin.pruneComplete(now2, expectedPruneUpperBound2);
+    } finally {
+      transactionPruningPlugin.destroy();
+      hBaseAdmin.disableTable(txEmptyTable);
+      hBaseAdmin.deleteTable(txEmptyTable);
+    }
+    }
```
gokulavasan: Fix alignment

gokulavasan:
Just one minor comment. LGTM 👍

PR #35, opened 2017-02-12
(TEPHRA-219) Execute cross region calls in Coprocessor as the login user

gokulavasan:
i) Fixes a problem where the pruneThread would exit if the state of the service is not yet set to RUNNING state. This also fixes the flakiness in the PruneUpperBoundSupplierTest which was caused due to the above problem.

ii) Load the txMaxLifeTimeMillis and pruneEnable properties dynamically.

iii) Add a hook around cross region calls in the PruneUpperBoundWriter.

### DIFF
--- a/tephra-hbase-compat-1.1-base/src/main/java/org/apache/tephra/hbase/coprocessor/TransactionProcessor.java
+++ b/tephra-hbase-compat-1.1-base/src/main/java/org/apache/tephra/hbase/coprocessor/TransactionProcessor.java
@@ -107,12 +107,13 @@ public class TransactionProcessor extends BaseRegionObserver {
 
   private final TransactionCodec txCodec;
   private TransactionStateCache cache;
-  private CompactionState compactionState;
+  private volatile CompactionState compactionState;
+
+  protected volatile Boolean pruneEnable;
+  protected volatile Long txMaxLifetimeMillis;
   protected Map<byte[], Long> ttlByFamily = Maps.newTreeMap(Bytes.BYTES_COMPARATOR);
   protected boolean allowEmptyValues = TxConstants.ALLOW_EMPTY_VALUES_DEFAULT;
   protected boolean readNonTxnData = TxConstants.DEFAULT_READ_NON_TX_DATA;
-  protected Long txMaxLifetimeMillis;
-  private Boolean pruneEnable;
 
   public TransactionProcessor() {
     this.txCodec = new TransactionCodec();
@@ -144,10 +145,12 @@ public class TransactionProcessor extends BaseRegionObserver {
       }
 
       this.allowEmptyValues = getAllowEmptyValues(env, tableDesc);
+      this.txMaxLifetimeMillis = getTxMaxLifetimeMillis(env);
       this.readNonTxnData = Boolean.valueOf(tableDesc.getValue(TxConstants.READ_NON_TX_DATA));
       if (readNonTxnData) {
         LOG.info("Reading pre-existing data enabled for table " + tableDesc.getNameAsString());
       }
+      initializePruneState(env);
     }
   }
 
@@ -171,9 +174,7 @@ public class TransactionProcessor extends BaseRegionObserver {
 
   @Override
   public void stop(CoprocessorEnvironment e) throws IOException {
-    if (compactionState != null) {
-      compactionState.stop();
-    }
+    resetPruneState();
   }
 
   @Override
@@ -246,6 +247,15 @@ public class TransactionProcessor extends BaseRegionObserver {
     // If the property is not present in the tableDescriptor, get it from the Configuration
     return  (allowEmptyValuesFromTableDesc != null) ?
       Boolean.valueOf(allowEmptyValuesFromTableDesc) : allowEmptyValuesFromConfig;
+  }
+
+  private long getTxMaxLifetimeMillis(RegionCoprocessorEnvironment env) {
+    Configuration conf = getConfiguration(env);
+    if (conf != null) {
+      return TimeUnit.SECONDS.toMillis(conf.getInt(TxConstants.Manager.CFG_TX_MAX_LIFETIME,
+                                                   TxConstants.Manager.DEFAULT_TX_MAX_LIFETIME));
+    }
+    return TimeUnit.SECONDS.toMillis(TxConstants.Manager.DEFAULT_TX_MAX_LIFETIME);
   }
 
   private boolean isFamilyDelete(List<Cell> familyCells) {
@@ -321,15 +331,10 @@ public class TransactionProcessor extends BaseRegionObserver {
     LOG.debug(String.format("Region %s: memstore size = %s, num store files = %s",
                             region.getRegionInfo().getRegionNameAsString(), memstoreSize, numStoreFiles));
     if (memstoreSize == 0 && numStoreFiles == 0) {
-      if (pruneEnable == null) {
-        initPruneState(e);
-      }
-
-      if (Boolean.TRUE.equals(pruneEnable)) {
+      if (compactionState != null) {
         compactionState.persistRegionEmpty(System.currentTimeMillis());
       }
     }
-
   }
 
   @Override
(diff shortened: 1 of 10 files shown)


tephra-hbase-compat-1.1-base/src/main/java/org/apache/tephra/hbase/txprune/PruneUpperBoundWriter.java
```diff
@@ -81,14 +84,23 @@ private void startFlushThread() {
     flushThread = new Thread("tephra-prune-upper-bound-writer") {
       @Override
       public void run() {
-        while ((!isInterrupted()) && isRunning()) {
+        Service.State serviceState = state();
+        while ((!isInterrupted()) &&
+          (serviceState.equals(Service.State.NEW) || serviceState.equals(Service.State.STARTING) ||
```
poornachandra: hmm - do you think it would simplify the code if we go back to using a `stop` flag?

tephra-hbase-compat-1.1-base/src/main/java/org/apache/tephra/hbase/txprune/PruneUpperBoundWriter.java
```diff
@@ -81,14 +84,23 @@ private void startFlushThread() {
...
+            serviceState.equals(Service.State.RUNNING))) {
           long now = System.currentTimeMillis();
           if (now > (lastChecked + pruneFlushInterval)) {
             // should flush data
             try {
               while (pruneEntries.firstEntry() != null) {
-                Map.Entry<byte[], Long> firstEntry = pruneEntries.firstEntry();
-                dataJanitorState.savePruneUpperBoundForRegion(firstEntry.getKey(), firstEntry.getValue());
+                final Map.Entry<byte[], Long> firstEntry = pruneEntries.firstEntry();
+                User.runAsLoginUser(new PrivilegedExceptionAction<Void>() {
```
poornachandra: Adding `User.runAsLoginUser` here brings in unnecessary security code into `PruneUpperBoundWriter` class. Also if we modify `TransactionProcessor` later to add other table operations, we may miss to wrap those calls. I think it would be better to do the wrapping in `TransactionProcessor` class itself. What do you say?

tephra-hbase-compat-1.1-base/src/main/java/org/apache/tephra/hbase/txprune/PruneUpperBoundWriter.java
gokulavasan: Yes might be simpler

gokulavasan:
@poornachandra The build is failing in a different compat module due to the PruneUpperBoundWriter issue that I have fixed in compat-1.1-base module. Once we port the changes to other modules, it should fix the build.

tephra-hbase-compat-1.1-base/src/main/java/org/apache/tephra/hbase/coprocessor/TransactionProcessor.java
```diff
@@ -454,28 +441,39 @@ protected Filter getTransactionFilter(Transaction tx, ScanType type, Filter filt
...
+
       if (Boolean.TRUE.equals(pruneEnable)) {
-        String pruneTable = conf.get(TxConstants.TransactionPruning.PRUNE_STATE_TABLE,
-                                     TxConstants.TransactionPruning.DEFAULT_PRUNE_STATE_TABLE);
-        long pruneFlushInterval = TimeUnit.SECONDS.toMillis(
-          conf.getLong(TxConstants.TransactionPruning.PRUNE_FLUSH_INTERVAL,
-                       TxConstants.TransactionPruning.DEFAULT_PRUNE_FLUSH_INTERVAL));
-        compactionState = new CompactionState(c.getEnvironment(), TableName.valueOf(pruneTable), pruneFlushInterval);
+        // pruneTable and pruneFlushInterval cannot be changed by simply loading the Configuration dynamically
+        // since we have only one flush thread across all regions and we might loose.
```
poornachandra: This comment can be removed now

tephra-hbase-compat-1.1-base/src/main/java/org/apache/tephra/hbase/coprocessor/TransactionProcessor.java
```diff
@@ -340,12 +345,7 @@ public InternalScanner preCompactScannerOpen(ObserverContext<RegionCoprocessorEn
     // Get the latest tx snapshot state for the compaction
     TransactionVisibilityState snapshot = cache.getLatestState();
 
-    if (pruneEnable == null) {
-      initPruneState(c);
-    }
-
     if (Boolean.TRUE.equals(pruneEnable)) {
-      // Record tx state before the compaction
```
poornachandra: This comment is still relevant, no?

tephra-hbase-compat-1.1-base/src/main/java/org/apache/tephra/hbase/coprocessor/TransactionProcessor.java
```diff
@@ -454,28 +441,39 @@ protected Filter getTransactionFilter(Transaction tx, ScanType type, Filter filt
...
+                                  pruneTable.getNameWithNamespaceInclAsString()));
         }
       }
     }
   }
 
+  protected void resetPruneState() {
+    pruneEnable = false;
+    if (compactionState != null) {
+      compactionState.stop();
```
poornachandra: It would be good to reset `compactionState` to `null` here.

tephra-hbase-compat-1.1-base/src/main/java/org/apache/tephra/hbase/coprocessor/TransactionProcessor.java
```diff
@@ -356,8 +356,8 @@ public InternalScanner preCompactScannerOpen(ObserverContext<RegionCoprocessorEn
   @Override
   public void postCompact(ObserverContext<RegionCoprocessorEnvironment> e, Store store, StoreFile resultFile,
                           CompactionRequest request) throws IOException {
-    // Persist the compaction state after a succesful compaction
-    if (compactionState != null) {
+    // Persist the compaction state after a successful compaction
+    if (Boolean.TRUE.equals(pruneEnable)) {
```
poornachandra: It would be better to change this check to `compactionState != null`. Otherwise there is a race condition during `initializePruneState()`, where `pruneEnable` could be `true`, but `compactionState` is still not initialized.

tephra-hbase-compat-1.1-base/src/main/java/org/apache/tephra/hbase/coprocessor/TransactionProcessor.java
```diff
@@ -454,28 +442,38 @@ protected Filter getTransactionFilter(Transaction tx, ScanType type, Filter filt
     return TransactionFilters.getVisibilityFilter(tx, ttlByFamily, allowEmptyValues, type, filter);
   }
 
-  private void initPruneState(ObserverContext<RegionCoprocessorEnvironment> c) {
-    Configuration conf = getConfiguration(c.getEnvironment());
-    // Configuration won't be null in TransactionProcessor but the derived classes might return
-    // null if it is not available temporarily
+  protected void initializePruneState(RegionCoprocessorEnvironment env) {
```
poornachandra: Good to add javadoc for the protected method

tephra-hbase-compat-1.1-base/src/main/java/org/apache/tephra/hbase/txprune/PruneUpperBoundWriter.java
```diff
@@ -97,27 +101,33 @@ private void startFlushThread() {
...
-              }
-              // Record empty regions
-              while (!emptyRegions.isEmpty()) {
-                Map.Entry<byte[], Long> firstEntry = emptyRegions.firstEntry();
-                dataJanitorState.saveEmptyRegionForTime(firstEntry.getValue(), firstEntry.getKey());
-                // We can now remove the entry only if the key and value match with what we wrote since it is
-                // possible that a new value for the same key has been added
-                emptyRegions.remove(firstEntry.getKey(), firstEntry.getValue());
-              }
+              User.runAsLoginUser(new PrivilegedExceptionAction<Void>() {
```
poornachandra: Let's not close TEPHRA-219 until we have figured out how to handle the wrapping cleanly.

poornachandra:
LGTM

PR #37, opened 2017-02-21
TEPHRA-223 Encapsulate the two data structures used for invalid transactions to avoid update issues

poornachandra:
JIRA - https://issues.apache.org/jira/browse/TEPHRA-223

Approach:
Encapsulate the invalid list and the invalidArray data structures into a class InvalidTxList. InvalidTxList tracks the changes to invalid list, and then populates invalidArray lazily when the list changes.

TODO: Add tests for class InvalidTxList

### DIFF
--- a/tephra-core/src/main/java/org/apache/tephra/TransactionManager.java
+++ b/tephra-core/src/main/java/org/apache/tephra/TransactionManager.java
@@ -29,7 +29,11 @@ import com.google.common.collect.Sets;
 import com.google.common.util.concurrent.AbstractService;
 import com.google.inject.Inject;
 import it.unimi.dsi.fastutil.longs.LongArrayList;
+import it.unimi.dsi.fastutil.longs.LongArraySet;
+import it.unimi.dsi.fastutil.longs.LongIterator;
+import it.unimi.dsi.fastutil.longs.LongSet;
 import org.apache.hadoop.conf.Configuration;
+import org.apache.tephra.manager.InvalidTxList;
 import org.apache.tephra.metrics.DefaultMetricsCollector;
 import org.apache.tephra.metrics.MetricsCollector;
 import org.apache.tephra.persist.NoOpTransactionStateStorage;
@@ -46,7 +50,6 @@ import org.slf4j.LoggerFactory;
 import java.io.IOException;
 import java.io.OutputStream;
 import java.util.Collection;
-import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.NavigableMap;
@@ -115,15 +118,11 @@ public class TransactionManager extends AbstractService {
   //poll every 10 second to emit metrics
   private static final long METRICS_POLL_INTERVAL = 10000L;
 
-  private static final long[] NO_INVALID_TX = { };
-
   // Transactions that are in progress, with their info.
   private final NavigableMap<Long, InProgressTx> inProgress = new ConcurrentSkipListMap<Long, InProgressTx>();
 
   // the list of transactions that are invalid (not properly committed/aborted, or timed out)
-  // TODO: explain usage of two arrays
-  private final LongArrayList invalid = new LongArrayList();
-  private long[] invalidArray = NO_INVALID_TX;
+  private final InvalidTxList invalidTxList = new InvalidTxList();
 
   // todo: use moving array instead (use Long2ObjectMap<byte[]> in fastutil)
   // todo: should this be consolidated with inProgress?
@@ -198,8 +197,7 @@ public class TransactionManager extends AbstractService {
   }
 
   private void clear() {
-    invalid.clear();
-    invalidArray = NO_INVALID_TX;
+    invalidTxList.clear();
     inProgress.clear();
     committedChangeSets.clear();
     committingChangeSets.clear();
@@ -318,7 +316,7 @@ public class TransactionManager extends AbstractService {
         txMetricsCollector.gauge("committing.size", committingChangeSets.size());
         txMetricsCollector.gauge("committed.size", committedChangeSets.size());
         txMetricsCollector.gauge("inprogress.size", inProgress.size());
-        txMetricsCollector.gauge("invalid.size", invalidArray.length);
+        txMetricsCollector.gauge("invalid.size", getInvalidSize());
       }
 
       @Override
@@ -327,7 +325,7 @@ public class TransactionManager extends AbstractService {
         txMetricsCollector.gauge("committing.size", committingChangeSets.size());
         txMetricsCollector.gauge("committed.size", committedChangeSets.size());
         txMetricsCollector.gauge("inprogress.size", inProgress.size());
-        txMetricsCollector.gauge("invalid.size", invalidArray.length);
+        txMetricsCollector.gauge("invalid.size", getInvalidSize());
       }
 
       @Override
@@ -363,7 +361,7 @@ public class TransactionManager extends AbstractService {
         }
         if (!timedOut.isEmpty()) {
           invalidEdits = Lists.newArrayListWithCapacity(timedOut.size());
-          invalid.addAll(timedOut.keySet());
+          invalidTxList.addAll(timedOut.keySet());
           for (Map.Entry<Long, InProgressType> tx : timedOut.entrySet()) {
             inProgress.remove(tx.getKey());
             // checkpoints never go into the committing change sets or the edits
@@ -373,9 +371,6 @@ public class TransactionManager extends AbstractService {
             }
           }
 
-          // todo: find a more efficient way to keep this sorted. Could it just be an array?
-          Collections.sort(invalid);
-          invalidArray = invalid.toLongArray();
           LOG.info("Invalidated {} transactions due to timeout.", timedOut.size());
         }
       }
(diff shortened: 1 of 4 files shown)


tephra-core/src/main/java/org/apache/tephra/manager/InvalidTxList.java
```diff
@@ -0,0 +1,110 @@
...
+    if (dirty) {
+      Collections.sort(invalid);
+      invalidArray = invalid.toLongArray();
+      dirty = false;
+    }
+    return invalidArray;
+  }
+
+  public List<Long> toList() {
+    return Collections.unmodifiableList(invalid);
```
gokulavasan: This won't be sorted right? But the comment on line 1114 of TransactionManager seems to assume that this method return a sorted list?

poornachandra: Good catch. I'll update InvalidTxList to lazily sort both the list and the array.

tephra-core/src/main/java/org/apache/tephra/manager/InvalidTxList.java
```diff
@@ -0,0 +1,110 @@
...
+    return invalid.size();
+  }
+
+  public boolean isEmpty() {
+    return invalid.isEmpty();
+  }
+
+  public boolean add(long id) {
+    boolean changed = invalid.add(id);
+    if (changed && !dirty) {
```
anew: dirty = dirty || changed

tephra-core/src/main/java/org/apache/tephra/manager/InvalidTxList.java
```diff
@@ -0,0 +1,110 @@
...
+      dirty = true;
+    }
+    return changed;
+  }
+
+  public boolean addAll(Collection<? extends Long> ids) {
+    boolean changed = invalid.addAll(ids);
+    if (changed && !dirty) {
```
anew: ditto

tephra-core/src/main/java/org/apache/tephra/manager/InvalidTxList.java
```diff
@@ -0,0 +1,110 @@
...
+    return changed;
+  }
+
+  public boolean contains(long id) {
+    return invalid.contains(id);
+  }
+
+  public boolean remove(long id) {
+    boolean changed = invalid.rem(id);
+    if (changed && !dirty) {
```
anew: ditto

tephra-core/src/main/java/org/apache/tephra/manager/InvalidTxList.java
```diff
@@ -0,0 +1,110 @@
...
+      dirty = true;
+    }
+    return changed;
+  }
+
+  public boolean removeAll(Collection<? extends Long> ids) {
+    boolean changed = invalid.removeAll(ids);
+    if (changed && !dirty) {
```
anew: ditto

tephra-core/src/main/java/org/apache/tephra/manager/InvalidTxList.java
anew: Or the caller can remove the assumption. Sorting takes n log n. Whereas the caller traverses the list once, which is n. Not sure it it worth sorting.

tephra-core/src/main/java/org/apache/tephra/manager/InvalidTxList.java
poornachandra: Thanks, it makes the code more readable

tephra-core/src/main/java/org/apache/tephra/manager/InvalidTxList.java
poornachandra: The sorting behavior is consistent with the existing implementation. Also truncate and snapshot (the two places where toList() is used) are not invoked frequently. So I suggest that, we'll leave this behavior as-is for now. We can take a look at this again when working on the performance improvements.

poornachandra:
> Does the added synchronization add performance overhead?

Extra synchronization is added in five places in this PR -
1. canCommit() - when transaction is not in in-progress  list

2. Emitting metrics - metrics are emitted every 10 seconds

3. getInvalidSize() - called when Thrift endpoint is invoked by user or by monitoring scripts

4. and 5. getExcludedListSize() and logStatistics() - these methods are used in test or unused

The only significant usage is by canCommit() when the committing transaction is not in the in-progress list. I think this overhead is acceptable since we now sort the invalid list only when creating a transaction, as opposed to sorting it whenever the list changed (like we did earlier).

poornachandra:
@anew @gokulavasan I have addressed the comments, please take another look

tephra-core/src/main/java/org/apache/tephra/manager/InvalidTxList.java
```diff
@@ -0,0 +1,113 @@
...
+    return invalid.isEmpty();
+  }
+
+  public boolean add(long id) {
+    boolean changed = invalid.add(id);
+    dirty = dirty || changed;
+    return changed;
+  }
+
+  public boolean addAll(Collection<? extends Long> ids) {
```
chtyim: the only call to this method is adding all child write pointers. It's better to take a `LongList` as the input so that the more efficient `invalid.addAll(LongList)` method can be called to avoid boxing (`LongArrayList` has a special implementation that use array copy instead of iterate + box one by one).

tephra-core/src/main/java/org/apache/tephra/TransactionManager.java
```diff
@@ -1124,7 +1110,7 @@ private boolean doTruncateInvalidTxBefore(long time) throws InvalidTruncateTimeE
     
     // Find all invalid transactions earlier than truncateWp
     Set<Long> toTruncate = Sets.newHashSet();
```
chtyim: Actually this can be a `LongSet` so that it can be passed to the `doTruncateInvalidTx` and eventually to `invalidTxList.removeAll` as a `LongSet` to have more efficient removal.

tephra-core/src/main/java/org/apache/tephra/manager/InvalidTxList.java
```diff
@@ -0,0 +1,113 @@
...
+ * The array is used to avoid creating a new array every time method {@link #toArray()} is invoked.
+ *
+ * This class is not thread safe and relies on external synchronization. TransactionManager always
+ * accesses an instance of this class after synchronization.
+ */
+@NotThreadSafe
+public class InvalidTxList {
+  private static final long[] NO_INVALID_TX = { };
+
+  private final LongArrayList invalid = new LongArrayList();
```
chtyim: Better have the type as `LongList`.

tephra-core/src/main/java/org/apache/tephra/manager/InvalidTxList.java
```diff
@@ -0,0 +1,113 @@
...
+   * @return sorted list of invalid transactions
+   */
+  public List<Long> toList() {
+    lazyUpdate();
+    return Collections.unmodifiableList(invalid);
+  }
+
+  private void lazyUpdate() {
+    if (dirty) {
+      Collections.sort(invalid);
```
chtyim: Actually there is no requirement in sorting the invalid `LongList` and it is more efficient to sort the `invalidArray` directly using `Arrays.sort(long[])` method.

`Collections.sort` first turn the list to an `Object[]` (hence boxing), sort the array, then set the value back to the list one by one.

tephra-core/src/main/java/org/apache/tephra/manager/InvalidTxList.java
poornachandra: Done

poornachandra:
@chtyim I have addressed the comments, please take a look

tephra-core/src/main/java/org/apache/tephra/manager/InvalidTxList.java
```diff
@@ -0,0 +1,127 @@
...
+    return changed;
+  }
+
+  public boolean addAll(Collection<? extends Long> ids) {
+    boolean changed = invalid.addAll(ids);
+    dirty = dirty || changed;
+    return changed;
+  }
+
+  public boolean addAll(LongCollection ids) {
```
chtyim: Should use `LongList` instead of `LongCollection` as the `LongArrayList` only optimize if it is a `LongList`.

tephra-core/src/main/java/org/apache/tephra/TransactionManager.java
```diff
@@ -1123,15 +1112,14 @@ private boolean doTruncateInvalidTxBefore(long time) throws InvalidTruncateTimeE
     }
     
     // Find all invalid transactions earlier than truncateWp
-    Set<Long> toTruncate = Sets.newHashSet();
-    for (long wp : invalid) {
-      // invalid list is sorted, hence can stop as soon as we reach a wp >= truncateWp
-      if (wp >= truncateWp) {
-        break;
+    LongSet toTruncate = new LongArraySet();
+    for (long wp : invalidTxList.toRawList()) {
```
chtyim: Shouldn't this use `toSortedArray`??

poornachandra: I reversed the check and iterate through the whole invalid list now, so sorting is no longer necessary.

chtyim: Unfortunately using the normal for-each loop means there would be boxing. To avoid boxing, you have to use the iterator directly

```java
LongIterator it = invalidTxList.toRawList().iterator();
while (it.hasNext()) {
  long wp = it.nextLong();
  ...
}
```

poornachandra: Good point

chtyim:
LGTM

PR #38, opened 2017-02-22
TEPHRA-224 Handle delay between transaction max lifetime check and data writes while pruning

poornachandra:
JIRA - https://issues.apache.org/jira/browse/TEPHRA-224

### DIFF
--- a/tephra-core/src/main/java/org/apache/tephra/TxConstants.java
+++ b/tephra-core/src/main/java/org/apache/tephra/TxConstants.java
@@ -376,6 +376,11 @@ public class TxConstants {
     public static final String PRUNE_FLUSH_INTERVAL = "data.tx.prune.flush.interval";
 
     /**
+     * The time in seconds used to pad transaction max lifetime while pruning.
+     */
+    public static final String PRUNE_GRACE_PERIOD = "data.tx.grace.period";
+
+    /**
      * Comma separated list of invalid transaction pruning plugins to load
      */
     public static final String PLUGINS = "data.tx.prune.plugins";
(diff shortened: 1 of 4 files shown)


tephra-core/src/main/java/org/apache/tephra/TxConstants.java
```diff
@@ -376,6 +376,11 @@
     public static final String PRUNE_FLUSH_INTERVAL = "data.tx.prune.flush.interval";
 
     /**
+     * The buffer in seconds used to pad transaction max lifetime while pruning.
+     */
+    public static final String PRUNE_BUFFER_INTERVAL = "data.tx.prune.buffer";
```
anew: it's not really an interval, or is it?

tephra-core/src/main/java/org/apache/tephra/txprune/TransactionPruningRunnable.java
```diff
@@ -57,8 +60,13 @@ public void run() {
       Transaction tx = txManager.startShort();
       txManager.abort(tx);
 
+      if (tx.getInvalids().length == 0) {
+        LOG.info("Invalid list is empty, not running transaction pruning");
```
anew: is this correct? Doesn't pruning also remove stale/out-of-date entries written by the plugins? Doesn't that cleanup have to happen regardless of whether we actually prune the invalid list?

tephra-core/src/main/java/org/apache/tephra/TxConstants.java
poornachandra: What would be a good name? `PRUNE_BUFFER_TIME`?

tephra-core/src/main/java/org/apache/tephra/txprune/TransactionPruningRunnable.java
poornachandra: Yes - I was planning on making that change later since that needs some analysis. I'll file a JIRA for that. Is that okay?

anew: so should we still prune here? It won't remove any invalid tx, but it will cleanup the prune state.

tephra-core/src/main/java/org/apache/tephra/TxConstants.java
anew: PRUNE_SAFETY_BUFFER? PRUNE_GRACE_PERIOD?

chtyim: grace period sounds good.

tephra-core/src/main/java/org/apache/tephra/txprune/TransactionPruningRunnable.java
poornachandra: I have filed TEPHRA-225 for to handle this

poornachandra:
@anew @chtyim  I have addressed comments. Please take another look.

anew:
LGTM

poornachandra:
Merged this PR to branch release/0.11.0-incubating

PR #39, opened 2017-03-01
(TEPHRA-226) Use older API to run as login user

gokulavasan:
JIRA : https://issues.apache.org/jira/browse/TEPHRA-226


poornachandra:
LGTM

PR #40, opened 2017-03-01
TEPHRA-227 Add a new command to get the set of regions that have not been compacted

gokulavasan:
Also adding a CLI command.

JIRA : https://issues.apache.org/jira/browse/TEPHRA-227


poornachandra:
LGTM

PR #41, opened 2017-03-23
TEPHRA-152 Using ReferenceCounting for TransactionStateCache refresh …

gokulavasan:
…thread, so that it can be stopped


poornachandra:
@gokulavasan Now that we have introduced class `ReferenceCountedSupplier`, it would be good to use this class in `PruneUpperBoundWriterSupplier` too. It will reduce code duplication.

PR #42, opened 2017-05-16
TEPHRA-228 Adding the ability to pass-in a clientId during the start …

gokulavasan:
…of a transaction which is logged when the transaction gets invalidated during time out.

### DIFF
--- a/tephra-core/src/main/java/org/apache/tephra/TransactionManager.java
+++ b/tephra-core/src/main/java/org/apache/tephra/TransactionManager.java
@@ -118,8 +118,11 @@ public class TransactionManager extends AbstractService {
   //poll every 10 second to emit metrics
   private static final long METRICS_POLL_INTERVAL = 10000L;
 
+  //Client id that is used if a client doesn't provide one while starting a transaction.
+  private static final String DEFAULT_CLIENTID = "unknown";
+
   // Transactions that are in progress, with their info.
-  private final NavigableMap<Long, InProgressTx> inProgress = new ConcurrentSkipListMap<Long, InProgressTx>();
+  private final NavigableMap<Long, InProgressTx> inProgress = new ConcurrentSkipListMap<>();
 
   // the list of transactions that are invalid (not properly committed/aborted, or timed out)
   private final InvalidTxList invalidTxList = new InvalidTxList();
@@ -127,8 +130,7 @@ public class TransactionManager extends AbstractService {
   // todo: use moving array instead (use Long2ObjectMap<byte[]> in fastutil)
   // todo: should this be consolidated with inProgress?
   // commit time next writePointer -> changes made by this tx
-  private final NavigableMap<Long, Set<ChangeId>> committedChangeSets =
-    new ConcurrentSkipListMap<Long, Set<ChangeId>>();
+  private final NavigableMap<Long, Set<ChangeId>> committedChangeSets = new ConcurrentSkipListMap<>();
   // not committed yet
   private final Map<Long, Set<ChangeId>> committingChangeSets = Maps.newConcurrentMap();
 
@@ -347,15 +349,16 @@ public class TransactionManager extends AbstractService {
         long currentTime = System.currentTimeMillis();
         Map<Long, InProgressType> timedOut = Maps.newHashMap();
         for (Map.Entry<Long, InProgressTx> tx : inProgress.entrySet()) {
-          long expiration = tx.getValue().getExpiration();
+          InProgressTx inProgressTx = tx.getValue();
+          long expiration = inProgressTx.getExpiration();
           if (expiration >= 0L && currentTime > expiration) {
             // timed out, remember tx id (can't remove while iterating over entries)
-            timedOut.put(tx.getKey(), tx.getValue().getType());
-            LOG.info("Tx invalid list: added tx {} because of timeout", tx.getKey());
+            timedOut.put(tx.getKey(), inProgressTx.getType());
+            LOG.info("Tx invalid list: added tx {} belonging to client '{}' because of timeout.",
+                     tx.getKey(), inProgressTx.getClientId());
           } else if (expiration < 0) {
             LOG.warn("Transaction {} has negative expiration time {}. Likely cause is the transaction was not " +
-                       "migrated correctly, this transaction will be expired immediately",
-                     tx.getKey(), expiration);
+                       "migrated correctly, this transaction will be expired immediately", tx.getKey(), expiration);
             timedOut.put(tx.getKey(), InProgressType.LONG);
           }
         }
@@ -592,8 +595,8 @@ public class TransactionManager extends AbstractService {
               } else if (type == null) {
                 type = TransactionType.SHORT;
               }
-              addInProgressAndAdvance(edit.getWritePointer(), edit.getVisibilityUpperBound(),
-                                      expiration, type);
+              // We don't persist the client id.
+              addInProgressAndAdvance(edit.getWritePointer(), edit.getVisibilityUpperBound(), expiration, type, null);
               break;
             case COMMITTING:
               addCommittingChangeSet(edit.getWritePointer(), edit.getChanges());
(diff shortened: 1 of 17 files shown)


tephra-core/src/main/java/org/apache/tephra/TransactionManager.java
```diff
@@ -763,15 +781,22 @@ private long getNextWritePointer() {
    * transaction moves it to the invalid list because we assume that its writes cannot be rolled back.
    */
   public Transaction startLong() {
+    return startLong(null);
+  }
+
+  /**
+   * Starts a long transaction with a client id.
+   */
+  public Transaction startLong(String clientId) {
```
gokulavasan: If we are going to deprecate the clientId-less methods, should we allow null here?

tephra-core/src/main/java/org/apache/tephra/TransactionManager.java
```diff
@@ -1378,17 +1403,32 @@ public TransactionType getTransactionType() {
     private final long expiration;
     private final InProgressType type;
     private final LongArrayList checkpointWritePointers;
+    private final String clientId;
```
gokulavasan: Not sure if we can add it to equals/hashCode. The unit tests fail when trying to compare the state before persistence and after recovery and won't be the same if we include clientId in comparison. Any reason why we should definitely include it? Or should we include it and adjust the test that does the comparison?

tephra-core/src/main/java/org/apache/tephra/TransactionManager.java
gokulavasan: Should I add Nullable annotation if we decide to allow null for clientId?

poornachandra: Eventually we want all client's to pass in a client i, hence the deprecation. If we allow nulls then that defeats the purpose, no?

gokulavasan: @poornachandra Sure sounds good. I will add a Precondition check for String not being null.

gokulavasan:
@poornachandra Addressed comments. Please take a look.

tephra-core/src/main/java/org/apache/tephra/TransactionSystemClient.java
```diff
@@ -44,11 +44,33 @@
   Transaction startShort(int timeout);
```
gokulavasan: Done.

tephra-core/src/main/java/org/apache/tephra/TransactionManager.java
gokulavasan: Added Preconditions null check for clientId.

tephra-core/src/main/java/org/apache/tephra/TransactionManager.java
poornachandra: It is okay not to add it now since we are not persisting it. Add a comment in the code.

tephra-hbase-compat-0.98/src/main/java/org/apache/tephra/hbase/coprocessor/TransactionProcessor.java
```diff
@@ -127,7 +126,7 @@ public TransactionProcessor() {
   public void start(CoprocessorEnvironment e) throws IOException {
     if (e instanceof RegionCoprocessorEnvironment) {
       RegionCoprocessorEnvironment env = (RegionCoprocessorEnvironment) e;
-      Supplier<TransactionStateCache> cacheSupplier = getTransactionStateCacheSupplier(env);
+      this.cacheSupplier = getTransactionStateCacheSupplier(env);
```
poornachandra: Unrelated change?

gokulavasan: Missed this change in https://github.com/apache/incubator-tephra/pull/41 for this compat-module. Hence fixing it here.

gokulavasan:
@poornachandra Can you please review the last commit. I have addressed your comment about TransactionContext change (deprecated old constructors and introduced a new one with the clientId)

gokulavasan:
@poornachandra Injecting ClientId in TransactionServiceClient now instead of changing the API of TransactionSystemClient. Please review when you get a chance.

gokulavasan:
@poornachandra Please review the last commit. I have addressed your comment about including hostname-pid as the default client id.

tephra-core/src/main/java/org/apache/tephra/runtime/ClientIdAwareTransactionModule.java
```diff
@@ -0,0 +1,41 @@
...
+
+import com.google.inject.AbstractModule;
+import com.google.inject.name.Names;
+import org.apache.tephra.TransactionSystemClient;
+import org.apache.tephra.distributed.TransactionServiceClient;
+
+/**
+ * Guice module that binds the clientId to be used in {@link TransactionSystemClient}.
+ */
+public abstract class ClientIdAwareTransactionModule extends AbstractModule {
```
poornachandra: Having `ClientIdAwareTransactionModule` as the parent class of Transaction modules does not add much value. Let's move the binding to the child modules.

tephra-core/src/main/java/org/apache/tephra/runtime/TransactionDistributedModule.java
```diff
@@ -35,13 +34,24 @@
...
+import java.lang.management.ManagementFactory;
+
 /**
  * Guice bindings for running in distributed mode on a cluster.
  */
-final class TransactionDistributedModule extends AbstractModule {
+final class TransactionDistributedModule extends ClientIdAwareTransactionModule {
+
+  public TransactionDistributedModule() {
+    super(ManagementFactory.getRuntimeMXBean().getName());
```
poornachandra: This default is already defined in `TransactionModules`. Do we still need it here? 
Same for `TransactionInMemoryModule` and `TransactionLocalModule`.

gokulavasan: @poornachandra There is no guarantee that TransactionModules().getDistributedModule() is used to get an instance of TransactionDistributedModule. One can just create an instance of this class.

tephra-core/src/main/java/org/apache/tephra/runtime/ClientIdAwareTransactionModule.java
gokulavasan: Sure ok.

tephra-core/src/main/java/org/apache/tephra/runtime/TransactionDistributedModule.java
poornachandra: In that case they can provide the client id since the default constructor won't be present, right?

gokulavasan:
@poornachandra Thanks for the comments. Addressed them.

tephra-core/src/main/java/org/apache/tephra/distributed/TransactionServiceClient.java
```diff
@@ -130,10 +134,11 @@ public static void doMain(boolean verbose, Configuration conf) throws Exception
    * for service discovery. Otherwise it will look for the port in the
    * config and use localhost.
    * @param config a configuration containing the zookeeper properties
+   * @param clientId id of the client that identifies it when it starts a transaction
    */
   @Inject
-  public TransactionServiceClient(Configuration config,
-                                  ThriftClientProvider clientProvider) {
+  public TransactionServiceClient(Configuration config, ThriftClientProvider clientProvider,
```
poornachandra: Some clients like Apache Phoenix do not use the guice modules. They instantiate `TransactionServiceClient` directly. It will be good to have the existing constructor too and assign the default value for the clientId in it.

gokulavasan: @poornachandra Addressed comment in the commit https://github.com/apache/incubator-tephra/pull/42/commits/0279658b3b0718298ccfe363343bce182e4ba655. Please take a look again.

tephra-core/src/main/java/org/apache/tephra/distributed/TransactionServiceClient.java
```diff
@@ -125,15 +130,20 @@ public static void doMain(boolean verbose, Configuration conf) throws Exception
     }
   }
 
+  public TransactionServiceClient(Configuration config, ThriftClientProvider clientProvider) {
```
poornachandra: Please add javadocs for the public constructor.

PR #43, opened 2017-05-17
TEPHRA-230 Fix reading of saved regions during invalid pruning run

poornachandra:
JIRA - https://issues.apache.org/jira/browse/TEPHRA-230


gokulavasan:
LGTM 👍

PR #44, opened 2017-06-21
TEPHRA-235 Ensure that TransactionSnapshot always have a sorted invalid transaction list

anwar6953:
[TEPHRA-235](https://issues.apache.org/jira/browse/TEPHRA-235) Ensure that TransactionSnapshot always have a sorted invalid transaction list.

### DIFF
--- a/tephra-core/src/test/java/org/apache/tephra/snapshot/SnapshotCodecTest.java
+++ b/tephra-core/src/test/java/org/apache/tephra/snapshot/SnapshotCodecTest.java
@@ -20,8 +20,11 @@ package org.apache.tephra.snapshot;
 
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSortedMap;
+import com.google.common.collect.Iterables;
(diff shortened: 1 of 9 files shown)


tephra-core/src/test/java/org/apache/tephra/snapshot/SnapshotCodecTest.java
```diff
@@ -375,4 +386,17 @@ private void assertTransactionVisibilityStateEquals(TransactionVisibilityState e
...
   }
+
+  private TransactionSnapshot assertSorted(TransactionSnapshot txSnapshot) {
+    Collection<Long> invalidTxList = txSnapshot.getInvalid();
+    if (invalidTxList.isEmpty()) {
+      return txSnapshot;
+    }
+    Long previousLong = Iterables.get(invalidTxList, 0);
+    for (Long aLong : invalidTxList) {
+      Assert.assertTrue(aLong >= previousLong);
```
poornachandra: Would be good to say sort error here

PR #45, opened 2017-06-25
[TEPHRA-236] Changes to replace deprecated classes

bijugs:
Changes to replace HBaseAdmin, HTable and HTableDescriptor classes which were deprecated in HBase 1.0. WIP PR on ``tephra-hbase-compat-1.1-base`` for initial comments.


bijugs:
@poornachandra Did update the PR by removing the updated code. Since this is ``WIP`` will squash the commits once the reviews are complete. At this time ``Table`` class doesn't support ``flushCommits`` method and have raised a [ticket](https://issues.apache.org/jira/browse/HBASE-18360) to ``Hbase`` project.

bijugs:
Looks like ``flushCommits`` is not required since it [is called every time](https://github.com/apache/hbase/blob/rel/1.1.0/hbase-client/src/main/java/org/apache/hadoop/hbase/client/HTable.java#L1015-L1019) a ``put`` call is made from ``Table``.

poornachandra:
Thanks @bijugs, I'll continue the review

tephra-hbase-compat-1.1-base/src/test/java/org/apache/tephra/hbase/TransactionAwareHTableTest.java
```diff
@@ -180,6 +182,7 @@ public static void setupBeforeClass() throws Exception {
 
     testUtil.startMiniCluster();
     hBaseAdmin = testUtil.getHBaseAdmin();
+    conn = testUtil.getConnection();
```
poornachandra: It will be a good idea to close the connection object in the method `shutdownAfterClass()`

tephra-hbase-compat-1.1-base/src/main/java/org/apache/tephra/hbase/SecondaryIndexTable.java
```diff
@@ -56,17 +58,24 @@
...
         hBaseAdmin.createTable(new HTableDescriptor(secondaryIndexTableName));
       }
-      secondaryIndexHTable = new HTable(hTable.getConfiguration(), secondaryIndexTableName);
-    } catch (Exception e) {
+      secondaryIndexHTable = conn.getTable(secondaryIndexTableName);
+    } catch (Exception e) { 
       Throwables.propagate(e);
+    } finally {
+      if (conn != null) {
+         conn.close();
```
bijugs: This connection is local to the constructor and gets closed when the constructor execution is complete. Looks like we don't need to take any further actions in ``close()``. Please let me know otherwise.

tephra-hbase-compat-1.1-base/src/test/java/org/apache/tephra/hbase/TransactionAwareHTableTest.java
bijugs: Will update the code to ``close`` the connection.

tephra-hbase-compat-1.1-base/src/main/java/org/apache/tephra/hbase/SecondaryIndexTable.java
```diff
@@ -56,17 +58,24 @@
...
+    try { 
+      conn  = ConnectionFactory.createConnection(hTable.getConfiguration());
+      Admin hBaseAdmin = conn.getAdmin();
       if (!hBaseAdmin.tableExists(secondaryIndexTableName)) {
         hBaseAdmin.createTable(new HTableDescriptor(secondaryIndexTableName));
       }
-      secondaryIndexHTable = new HTable(hTable.getConfiguration(), secondaryIndexTableName);
-    } catch (Exception e) {
+      secondaryIndexHTable = conn.getTable(secondaryIndexTableName);
+    } catch (Exception e) {
```
bijugs: Not sure where the whitespaces are from. Probably we may have to use ``git apply --reject --whitespace=fix mychanges.path``?

tephra-hbase-compat-1.1-base/src/test/java/org/apache/tephra/hbase/TransactionAwareHTableTest.java
bijugs: @poornachandra new commit to close the ``connection`` added to the PR.

bijugs:
@poornachandra patch after squashing the commits seems to apply fine.

tephra-hbase-compat-1.1-base/src/main/java/org/apache/tephra/hbase/SecondaryIndexTable.java
poornachandra: @bijugs The connection created in the constructor has to be closed after the table is closed. Take a look at https://hbase.apache.org/1.1/apidocs/org/apache/hadoop/hbase/client/ConnectionFactory.html for an example.

tephra-hbase-compat-1.1-base/src/main/java/org/apache/tephra/hbase/SecondaryIndexTable.java
poornachandra: 👍

tephra-hbase-compat-1.1-base/src/test/java/org/apache/tephra/hbase/TransactionAwareHTableTest.java
poornachandra: 👍

poornachandra:
@bijugs Just one more comment to address - https://github.com/apache/incubator-tephra/pull/45#discussion_r128894170

bijugs:
@poornachandra did over look the logic and thank you for catching that. Updated the PR with an additional commit to make the review process easier. Can squash the commits once the review of the latest change is complete.

tephra-hbase-compat-1.1-base/src/main/java/org/apache/tephra/hbase/SecondaryIndexTable.java
```diff
@@ -177,11 +173,23 @@ public void close() throws IOException {
...
       } catch (IOException ex) {
         e.addSuppressed(e);
       }
       throw e;
     }
     secondaryIndexTable.close();
+    conn.close();
+  }
+
+  protected void finalize() throws Throwable {
```
poornachandra: We try not to rely on `finalize()` since there is no guarantee when it will be called. Also since we are closing the connection in the `close()` method, there is no need for finalize. Let's remove this method.

tephra-hbase-compat-1.1-base/src/main/java/org/apache/tephra/hbase/SecondaryIndexTable.java
```diff
@@ -177,11 +173,23 @@ public void close() throws IOException {
     } catch (IOException e) {
       try {
         secondaryIndexTable.close();
+        conn.close();
```
poornachandra: `conn` may not be closed if `secondaryIndexTable.close()` throws an exception.

Adding another nested try-catch will make the code complex. Instead what do you think about putting each close in its own try-catch block, remember the first exception, and add the other exceptions as suppressed to the first exception?

tephra-hbase-compat-1.1-base/src/main/java/org/apache/tephra/hbase/SecondaryIndexTable.java
bijugs: Was being overly cautious to not leave the connection open. Will remove it.

tephra-hbase-compat-1.1-base/src/main/java/org/apache/tephra/hbase/SecondaryIndexTable.java
bijugs: The ``finalize()`` would have helped in this scenario. Given that we don't want to use it as a standard, definitely we can go with either approach. Will make another commit with the changes.

bijugs:
@poornachandra , Updated the PR with changes for the latest comments and squashed the commits to one.

poornachandra:
Thanks for the contribution @bijugs!

PR #48, opened 2017-08-31
[TEPHRA-241] Introduce a way to limit the size of a transaction

anew:
- first commit introduces four new configurations for change set size limits, enforces these limits in a new method canCommitOrThrow() (we keep canCommit() for compatibility), and uses the new method throughout the higher level classes and tests
- second commit refactors the transaction system test cases: Don't start a new TxManager for each test method, have a shared common configuration across subclasses, move the checkpoint test to the base class such that it also gets tested with Thrift
- third commit adds a new test case for size limit validation

Thrift changes are backward-compatible (we keep the existing methods the same), and so are the TransactionSystemClient changes.

### DIFF
--- a/tephra-core/src/main/java/org/apache/tephra/distributed/thrift/TTransactionServer.java
+++ b/tephra-core/src/main/java/org/apache/tephra/distributed/thrift/TTransactionServer.java
@@ -68,6 +68,8 @@ public class TTransactionServer {
 
     public TBoolean canCommitTx(TTransaction tx, Set<ByteBuffer> changes) throws TTransactionNotInProgressException, org.apache.thrift.TException;
 
+    public TBoolean canCommitOrThrow(TTransaction tx, Set<ByteBuffer> changes) throws TTransactionNotInProgressException, TGenericException, org.apache.thrift.TException;
+
     public TBoolean commitTx(TTransaction tx) throws TTransactionNotInProgressException, org.apache.thrift.TException;
 
     public void abortTx(TTransaction tx) throws org.apache.thrift.TException;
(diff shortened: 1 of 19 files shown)


tephra-core/src/test/java/org/apache/tephra/TransactionManagerTest.java
```diff
@@ -54,115 +54,35 @@ protected TransactionStateStorage getStateStorage() {
...
   }
 
-  @After
-  public void after() {
+  @AfterClass
+  public static void afterClass() {
     txManager.stopAndWait();
   }
 
-  @Test
```
anew: this test moved to TransactionSystemTest without change.

tephra-core/src/main/java/org/apache/tephra/distributed/thrift/TTransactionServer.java
```diff
@@ -68,6 +68,8 @@
 
     public TBoolean canCommitTx(TTransaction tx, Set<ByteBuffer> changes) throws TTransactionNotInProgressException, org.apache.thrift.TException;
 
+    public TBoolean canCommitOrThrow(TTransaction tx, Set<ByteBuffer> changes) throws TTransactionNotInProgressException, TGenericException, org.apache.thrift.TException;
```
anew: please do not review this file, it is generated by Thrift. Only need to review transaction.thrift

anew:
Travis build failed again in the known flaky test with Java 8. It does pass when I run it with Java 8 locally. Java 7 also passes in Travis. I am confident that this was not introduced by the changes in this PR, and will commit this now.

anew:
After rebasing on latest master, another travis build is running; waiting for that to finish.

anew:
Same travis failure. This appears to happen a lot more frequently with Java 8. I will commit this now and try to fix the flaky test before 0.13 release.

PR #49, opened 2017-08-31
[TEPHRA-242] Ensure Pruning Service shuts down

anew:
Shutdown the pruning service before the tx server, to ensure it shuts down even if tx server shutdown is interrupted.

### DIFF
--- a/tephra-core/src/main/java/org/apache/tephra/distributed/TransactionService.java
+++ b/tephra-core/src/main/java/org/apache/tephra/distributed/TransactionService.java
@@ -121,14 +121,14 @@ public final class TransactionService extends InMemoryTransactionService {
       @Override
       public void follower() {
         ListenableFuture<State> stopFuture = null;
+        if (pruningService != null && pruningService.isRunning()) {
+          // Wait for pruning service to stop after un-registering from discovery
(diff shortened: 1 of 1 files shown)


gokulavasan:
LGTM 👍

tephra-core/src/main/java/org/apache/tephra/distributed/TransactionService.java
```diff
@@ -121,15 +121,15 @@ public void failed(State from, Throwable failure) {
...
         if (server != null && server.isRunning()) {
           server.stopAndWait();
         }
-        if (pruningService != null && pruningService.isRunning()) {
-          // Wait for pruning service to stop after un-registering from discovery
-          stopFuture = pruningService.stop();
-        }
         undoRegister();
 
         if (stopFuture != null) {
```
poornachandra: It would be good to add a timeout to the `Futures.getUnchecked(stopFuture)` in the next line, so that we don't block forever.

anew: pruningService.shutdown() itself has a 30 second timeout. So this future is guaranteed to return, and we don't need a timeout here.

PR #50, opened 2017-09-05
[TEPHRA-238] Support for HBase 1.3.x

anew:
- Adds a new compat module using 1.3.1 as the HBase version.  I could not reuse the compat-1.1-base because HTable added four new methods in 1.3 (get/setOperation/RpcTimeout)
- All the classes in the new src/ tree are the same as in the compat-1.1-base, except:
1. implemented the four new methods in TransactionAwareHTable
2. change to a TransactionProcessorTest.createRegion(), because WALFactory.getWAL() now takes an extra argument for the namespace. 
3. Changing 1.1 to 1.3 in all comments etc.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -157,6 +157,7 @@
     <module>tephra-hbase-compat-1.0</module>
     <module>tephra-hbase-compat-1.0-cdh</module>
     <module>tephra-hbase-compat-1.1-base</module>
+    <module>tephra-hbase-compat-1.3</module>
     <module>tephra-examples</module>
     <module>tephra-distribution</module>
   </modules>
@@ -212,6 +213,7 @@
     <hbase11.version>1.1.1</hbase11.version>
     <hbase12cdh.version>1.2.0-cdh5.7.0</hbase12cdh.version>
     <hbase12.version>1.2.0</hbase12.version>
+    <hbase13.version>1.3.1</hbase13.version>
     <junit.version>4.11</junit.version>
     <slf4j.version>1.7.5</slf4j.version>
     <thrift.version>0.9.0</thrift.version>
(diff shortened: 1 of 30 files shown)


JamesRTaylor:
+1. Thanks, @anew!

anew:
Thanks for the quick review, @JamesRTaylor. As an additional validation, could you run the Phoenix tests with this branch of Tephra? Thanks a lot!

JamesRTaylor:
Will do - I'll get back to you when finished.

anew:
Rebased on top of Tephra-252. Travis build should now pass.

JamesRTaylor:
Verified that all test pass with usage of new HBase 1.3 compat module using this patch. Would be much appreciated if a Tephra release could be done with this patch so we can support transactions in our HBase 1.3 release for 4.12.

anew:
Thanks, @JamesRTaylor, for the verification. I am waiting for the Travis build to succeed. 
What is your anticipated date for Phoenix 4.12? We have a few critical fixes that we would like to get done before the release, so a release in about 2 weeks seems realistic.

anew:
The travis test run for Java 8 failed while it passed for Java 7. It fails in a test for HBase 1.2, which is not affected at all by this change. That indicates that the failing test case is flaky. I created a new Jira TEPHRA-253 to track the flaky test. 

Since that failure is due to a flaky test, and it passed with Java 7, and the changes here do not have any influence on the failed test, I am confident that this change did not introduce the failure. Hence I will commit this.

JamesRTaylor:
The sooner the better for a Tephra release. This is the critical issue for us as we have no Tephra support for HBase 1.3 currently.

anew:
We also want to do a release soon. But we must fix at least TEPHRA-239, 242 and 244 as they are show stoppers for people who use Tephra in production. 244 is definitely the most complex to fix. Anyway this is not the place to have a release discussion... let's discuss the release on the mailing list when 244 is fixed? For now, this issue is resolved, thanks for the help.

PR #51, opened 2017-09-06
[TEPHRA-252] Fix Travis build for Oracle JDK 7

anew:
The issue is that Travis is upgrading/upgraded to Ubuntu Trusty which does not have/support Oracle JDK 7. The fix is, for now, to explicitly require the old distro, namely "precise".

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,9 @@
 
 language: java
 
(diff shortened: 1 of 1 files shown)


chtyim:
LGTM

anew:
Travis build passed now. Will commit this.

PR #52, opened 2017-09-08
(TEPHRA-250) Allow to trigger transaction pruning

anew:
- adds a method to TransactionPruningService to trigger pruning now
- exposes this method through Thrift and the Tx system client
- exposes the pruning service so that systems embedding Tephra can call the new method


tephra-core/src/main/java/org/apache/tephra/distributed/thrift/TTransactionServer.java
```diff
@@ -88,6 +88,8 @@
 
     public TTransaction checkpoint(TTransaction tx) throws TTransactionNotInProgressException, org.apache.thrift.TException;
```
anew: do not review this file, it is generated by Thrift.

PR #53, opened 2017-09-09
(TEPHRA-243) Improve logging for slow log append

anew:
- adds more detail to the log message (number of entries, bytes written)
- ensures only one thread logs the message concurrently, by moving it into the method actually writes to storage. Previously all threads logged it even if they did not write themselves 
- changed ThriftTransactionSystemTest to use LocalTransactionLog (otherwise it does not get tested at all)
- fixed a few unrelated compiler warning encountered on the way

### DIFF
--- a/tephra-core/src/main/java/org/apache/tephra/persist/AbstractTransactionLog.java
+++ b/tephra-core/src/main/java/org/apache/tephra/persist/AbstractTransactionLog.java
@@ -19,9 +19,11 @@
 package org.apache.tephra.persist;
 
 import com.google.common.annotations.VisibleForTesting;
-import com.google.common.collect.Lists;
+import com.google.common.base.Stopwatch;
+import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.io.LongWritable;
 import org.apache.hadoop.io.Writable;
+import org.apache.tephra.TxConstants;
 import org.apache.tephra.metrics.MetricsCollector;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -29,17 +31,20 @@ import org.slf4j.LoggerFactory;
 import java.io.DataInput;
 import java.io.DataOutput;
 import java.io.IOException;
-import java.util.LinkedList;
+import java.util.Collections;
 import java.util.List;
+import java.util.Queue;
+import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.atomic.AtomicLong;
+import javax.annotation.Nullable;
 
 /**
  * Common implementation of a transaction log, backed by file reader and writer based storage.  Classes extending
  * this class, must also implement {@link TransactionLogWriter} and {@link TransactionLogReader}.
+ *
+ * It is important to call close() on this class to ensure that all writes are synced and the log files are closed.
  */
 public abstract class AbstractTransactionLog implements TransactionLog {
-  /** Time limit, in milliseconds, of an append to the transaction log before we log it as "slow". */
-  private static final long SLOW_APPEND_THRESHOLD = 1000L;
 
   private static final Logger LOG = LoggerFactory.getLogger(AbstractTransactionLog.class);
 
@@ -47,22 +52,32 @@ public abstract class AbstractTransactionLog implements TransactionLog {
   private final MetricsCollector metricsCollector;
   protected long timestamp;
   private volatile boolean initialized;
+  private volatile boolean closing;
   private volatile boolean closed;
-  private AtomicLong syncedUpTo = new AtomicLong();
-  private List<Entry> pendingWrites = Lists.newLinkedList();
+  private long writtenUpTo = 0L;
+  private volatile long syncedUpTo = 0L;
+  private final Queue<Entry> pendingWrites = new ConcurrentLinkedQueue<>();
   private TransactionLogWriter writer;
 
-  public AbstractTransactionLog(long timestamp, MetricsCollector metricsCollector) {
+  private int countSinceLastSync = 0;
+  private long positionBeforeWrite = -1L;
+  private final Stopwatch stopWatch = new Stopwatch();
+
+  private final long slowAppendThreshold;
+
+  AbstractTransactionLog(long timestamp, MetricsCollector metricsCollector, Configuration conf) {
     this.timestamp = timestamp;
     this.metricsCollector = metricsCollector;
+    this.slowAppendThreshold = conf.getLong(TxConstants.TransactionLog.CFG_SLOW_APPEND_THRESHOLD,
+                                            TxConstants.TransactionLog.DEFAULT_SLOW_APPEND_THRESHOLD);
   }
 
   /**
-   * Initializes the log file, opening a file writer.  Clients calling {@code init()} should ensure that they
-   * also call {@link HDFSTransactionLog#close()}.
+   * Initializes the log file, opening a file writer.
+   *
    * @throws java.io.IOException If an error is encountered initializing the file writer.
    */
-  public synchronized void init() throws IOException {
+  private synchronized void init() throws IOException {
     if (initialized) {
       return;
     }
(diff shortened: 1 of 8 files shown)


tephra-core/src/main/java/org/apache/tephra/persist/AbstractTransactionLog.java
```diff
@@ -211,26 +212,36 @@ private void sync() throws IOException {
       // prevent writer being dereferenced
```
anew: I am not sure why this is needed. As long as this method is running, there is a stack frame (= a GC root) that has a reference to this, and hence to its members. Anybody know why this is here?

tephra-core/src/main/java/org/apache/tephra/persist/AbstractTransactionLog.java
```diff
@@ -48,21 +52,30 @@
...
   private volatile boolean initialized;
   private volatile boolean closed;
+  private AtomicLong writtenUpTo = new AtomicLong();
   private AtomicLong syncedUpTo = new AtomicLong();
-  private List<Entry> pendingWrites = Lists.newLinkedList();
+  private Queue<Entry> pendingWrites = new ConcurrentLinkedQueue<>();
   private TransactionLogWriter writer;
 
-  public AbstractTransactionLog(long timestamp, MetricsCollector metricsCollector) {
+  private final AtomicLong positionBeforeAppend = new AtomicLong(0);
```
chtyim: This can be just a `long` field, right? Since it is only used by the start/stop timer method, which only gets called inside the writer sync block.

tephra-core/src/main/java/org/apache/tephra/persist/AbstractTransactionLog.java
```diff
@@ -48,21 +52,30 @@
...
+  private final AtomicInteger countSinceLastSync = new AtomicInteger(0);
```
chtyim: And this can be just an `int` as well.

tephra-core/src/main/java/org/apache/tephra/persist/AbstractTransactionLog.java
chtyim: I don't think it is needed since writer only get assigned once in the `init` method. It was there since the beginning without any reason.

tephra-core/src/main/java/org/apache/tephra/persist/AbstractTransactionLog.java
```diff
@@ -165,26 +203,36 @@ private void sync() throws IOException {
...
-        entryCount++;
-        latestSeq = Math.max(latestSeq, e.getKey().get());
+      Entry[] currentPending = getPendingWrites();
+      if (currentPending != null) {
+        entryCount = currentPending.length;
+        startTimerIfNeeded(tmpWriter, entryCount);
+        tmpWriter.commitMarker(entryCount);
+        for (Entry e : currentPending) {
+          tmpWriter.append(e);
+          latestSeq = Math.max(latestSeq, e.getKey().get());
```
chtyim: Is the `Math.max` actually needed? The ordering of the sequence number is the always increasing and is maintained in the queue (hence the array here). I think we just need to have `latestSeq = currentPending[currentPending.length - 1].getKey().get()` after the loop, right?

tephra-core/src/main/java/org/apache/tephra/persist/AbstractTransactionLog.java
```diff
@@ -165,26 +203,36 @@ private void sync() throws IOException {
...
-      if (!currentPending.isEmpty()) {
-        tmpWriter.commitMarker(currentPending.size());
-      }
-
-      // write out all accumulated entries to log.
-      for (Entry e : currentPending) {
-        tmpWriter.append(e);
-        entryCount++;
-        latestSeq = Math.max(latestSeq, e.getKey().get());
+      Entry[] currentPending = getPendingWrites();
```
chtyim: Shall we move this statement out of the synchronize block and call it before the sync block and hold on to the `logSequence` object lock instead?

chtyim: hum... nevermind, seems like doesn't gain much.

tephra-core/src/main/java/org/apache/tephra/persist/AbstractTransactionLog.java
```diff
@@ -48,21 +52,30 @@
   protected long timestamp;
   private volatile boolean initialized;
   private volatile boolean closed;
+  private AtomicLong writtenUpTo = new AtomicLong();
```
chtyim: I think this can also be just a `long`.

tephra-core/src/main/java/org/apache/tephra/persist/AbstractTransactionLog.java
```diff
@@ -48,21 +52,30 @@
...
   private AtomicLong syncedUpTo = new AtomicLong();
```
chtyim: And this can be `long`.

tephra-core/src/main/java/org/apache/tephra/persist/AbstractTransactionLog.java
```diff
@@ -48,21 +52,30 @@
...
-  private List<Entry> pendingWrites = Lists.newLinkedList();
+  private Queue<Entry> pendingWrites = new ConcurrentLinkedQueue<>();
```
chtyim: final

tephra-core/src/main/java/org/apache/tephra/persist/AbstractTransactionLog.java
chtyim: Actually not. Just notice it is used outside of writer sync block.

chtyim: but, please make it `final`.

tephra-core/src/main/java/org/apache/tephra/persist/AbstractTransactionLog.java
```diff
@@ -165,26 +203,36 @@ private void sync() throws IOException {
...
+      Entry[] currentPending = getPendingWrites();
+      if (currentPending != null) {
+        entryCount = currentPending.length;
+        startTimerIfNeeded(tmpWriter, entryCount);
+        tmpWriter.commitMarker(entryCount);
+        for (Entry e : currentPending) {
+          tmpWriter.append(e);
+          latestSeq = Math.max(latestSeq, e.getKey().get());
+        }
+        writtenUpTo.set(latestSeq);
```
chtyim: We actually don't need this. Just have the `latestSeq` be a `volatile long` field, and have it updated here to the latest seq number in the current entry array. Then in the "sync" synchronize block below, the value can be set into `syncedUpTo`.

tephra-core/src/main/java/org/apache/tephra/persist/AbstractTransactionLog.java
chtyim: or a `volatile long` also work since we only use `get` and `set`.

chtyim:
I think the change is logically correct, just couple minor comments about simplifying the field types.

tephra-core/src/main/java/org/apache/tephra/persist/AbstractTransactionLog.java
anew: That's what i thought... removing it now.

tephra-core/src/main/java/org/apache/tephra/persist/AbstractTransactionLog.java
anew: oh yes, after the refactoring in the second commit it does not need to be atomic

tephra-core/src/main/java/org/apache/tephra/persist/AbstractTransactionLog.java
anew: yes

tephra-core/src/main/java/org/apache/tephra/persist/AbstractTransactionLog.java
anew: yes, actually that has always been the case, even before these changes...

tephra-core/src/main/java/org/apache/tephra/persist/AbstractTransactionLog.java
anew: yes it is only called within synchronized {}

tephra-core/src/main/java/org/apache/tephra/persist/AbstractTransactionLog.java
anew: good suggestion, done

tephra-core/src/main/java/org/apache/tephra/persist/AbstractTransactionLog.java
anew: I was thinking about that. However, it would add latency. In the following case: Suppose:
- thread 1 writes up to seq id 5
- thread 2 writes up to seq id 10
- thread 2 syncs and sets syncedUpto to 10
- thread 3 writes up to seq id 15
- now thread 1 checks whether it needs to sync. Because its latestSeq is 5, it can return. But if it compares with writtenUpto, then it would try to sync. 

That would add latency to thread 1 while thread 3 must wait anyway. That's why I introduced the writtenUpto

By the way, the existing code had syncedUpto wrong. It was always set to latestSeq of the thread that syncs. But it might have synced more than its own entries. For example, in the scenario above, if thread 1 syncs instead of thread 2, then it has synced everything up to 10, but it would set syncedUpto to 5. With the consequence that thread would sync again, needlessly.

anew:
Thanks for t he review, I agree with everything but your last comment. I have pushed another commit to address these comments.

tephra-core/src/main/java/org/apache/tephra/persist/AbstractTransactionLog.java
chtyim: Sounds good.

tephra-core/src/main/java/org/apache/tephra/persist/AbstractTransactionLog.java
```diff
@@ -48,21 +51,30 @@
...
+                                            TxConstants.TransactionLog.DEFAULT_SLOW_APPEND_THRESHOLD);
   }
 
   /**
    * Initializes the log file, opening a file writer.  Clients calling {@code init()} should ensure that they
    * also call {@link HDFSTransactionLog#close()}.
    * @throws java.io.IOException If an error is encountered initializing the file writer.
    */
-  public synchronized void init() throws IOException {
+  private synchronized void init() throws IOException {
```
chtyim: Seems like the javadoc is actually wrong. No "client" is calling this method and the contract should be whoever construct this class should call "close"?

chtyim:
Just one last comment about javadoc, rest LGTM.

tephra-core/src/main/java/org/apache/tephra/persist/AbstractTransactionLog.java
anew: yes, I noticed that, too, that's why I made it private, but forgot to update the Javadoc

chtyim:
LGTM

tephra-core/src/main/java/org/apache/tephra/persist/AbstractTransactionLog.java
```diff
@@ -134,57 +124,112 @@ private void ensureAvailable() throws IOException {
...
   }
 
   private void sync() throws IOException {
     // writes out pending entries to the HLog
-    TransactionLogWriter tmpWriter = null;
     long latestSeq = 0;
     int entryCount = 0;
     synchronized (this) {
       if (closed) {
         return;
```
poornachandra: It would be good if this throws an exception to indicate that the entries could not be synced. Silently returning will return success to the clients.

tephra-core/src/main/java/org/apache/tephra/persist/AbstractTransactionLog.java
```diff
@@ -85,44 +99,20 @@ public long getTimestamp() {
...
+    append(Collections.singletonList(edit));
   }
 
   @Override
   public void append(List<TransactionEdit> edits) throws IOException {
-    long startTime = System.nanoTime();
-    synchronized (this) {
+    // synchronizing here ensures that elements in the queue are ordered by seq number
+    synchronized (logSequence) {
       ensureAvailable();
```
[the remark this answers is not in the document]
anew: Good point, actually it is only used in one place, so I will just inline it outside the synchronize.

tephra-core/src/main/java/org/apache/tephra/persist/AbstractTransactionLog.java
anew: It would not always be correct to fail, because the edits from the current thread may actually have been synced. 

The logic in close() (which is also synchronized) is that this only gets closed after syncing. So a sync has just happened. However, it seems that it is possible that append() is called concurrently and it might append new edits, so there is a possible race condition. 

I guess we need an extra flag named closing. Close would first set closing to true. Now all calls to append() will fail. Then it will sync and set closed to true, such that any remaining threads do not attempt to sync.

tephra-core/src/main/java/org/apache/tephra/persist/AbstractTransactionLog.java
```diff
@@ -134,57 +124,112 @@ private void ensureAvailable() throws IOException {
...
+        if (syncedUpTo < latestSeq) {
+          writer.sync();
+          syncedUpTo = writtenUpTo;
+          stopTimerIfNeeded(writer);
+        }
+      }
     }
+    // in any case, emit metrics for the number entries we wrote.
+    // because the thread that actually syncs does not know how many it synced (it not write all of them)
+    metricsCollector.histogram("wal.sync.size", entryCount);
```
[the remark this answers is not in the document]
anew: we could emit that metric in stopTimerIfNeeded(). And actually this one, too.

tephra-core/src/main/java/org/apache/tephra/persist/AbstractTransactionLog.java
```diff
@@ -85,105 +100,145 @@ public long getTimestamp() {
...
-    long lastSynced = syncedUpTo.get();
+    // giving up the sync lock here allows other threads to write their edits before the sync happens.
+    // hence, we can have the edits from n threads in one sync.
+
     // someone else might have already synced our edits, avoid double syncing
-    if (lastSynced < latestSeq) {
-      tmpWriter.sync();
-      metricsCollector.histogram("wal.sync.size", entryCount);
-      syncedUpTo.compareAndSet(lastSynced, latestSeq);
+    if (syncedUpTo < latestSeq) {
```
poornachandra: Since this check is happening outside the synchronized block, `latestSeq` may not give the right value - we can just use `writtenUpTo` instead of `latestSeq`. 

To be safe, let's move the definition of both `latestSeq` and `entryCount` into the synchronized block on line 196, so that they cannot be used outside the block.

anew: Please see the previous discussion of this at https://github.com/apache/incubator-tephra/pull/53#pullrequestreview-62374698

PR #55, opened 2017-09-11
TEPHRA-244 Remove regions of deleted tables when computing prune upper bound

poornachandra:
JIRA - https://issues.apache.org/jira/browse/TEPHRA-244

This PR just removes regions of deleted tables when computing prune upper bound. This ensures that transient tables do not block pruning. However, if a transient table is created/deleted again with the same name and never gets a chance to be compacted, then it has a potential to block pruning.

Also the changes are made in one compact module only. After the review I'll port the changes to the other compact modules.

### DIFF
--- a/tephra-hbase-compat-1.1-base/src/main/java/org/apache/tephra/hbase/txprune/HBaseTransactionPruningPlugin.java
+++ b/tephra-hbase-compat-1.1-base/src/main/java/org/apache/tephra/hbase/txprune/HBaseTransactionPruningPlugin.java
@@ -20,6 +20,7 @@
 package org.apache.tephra.hbase.txprune;
 
 import com.google.common.base.Function;
+import com.google.common.base.Predicate;
 import com.google.common.collect.Iterables;
 import com.google.common.collect.Maps;
 import com.google.common.collect.Sets;
@@ -43,8 +44,10 @@ import org.slf4j.LoggerFactory;
 
 import java.io.IOException;
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.SortedSet;
 import java.util.TreeMap;
 import java.util.TreeSet;
@@ -286,6 +289,10 @@ public class HBaseTransactionPruningPlugin implements TransactionPruningPlugin {
    * @throws IOException when not able to talk to HBase
    */
   private long computePruneUpperBound(TimeRegions timeRegions) throws IOException {
+    // Get the tables for the current time from the latest regions set
+    final Set<TableName> existingTables = getTableNamesForRegions(timeRegions.getRegions());
+    LOG.debug("Tables for time {} = {}", timeRegions.getTime(), existingTables);
+
     do {
       LOG.debug("Computing prune upper bound for {}", timeRegions);
       SortedSet<byte[]> transactionalRegions = timeRegions.getRegions();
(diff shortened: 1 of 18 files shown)


tephra-hbase-compat-1.1-base/src/main/java/org/apache/tephra/hbase/txprune/HBaseTransactionPruningPlugin.java
```diff
@@ -335,6 +351,25 @@ private long computePruneUpperBound(TimeRegions timeRegions) throws IOException
     return -1;
   }
 
+  private SortedSet<byte[]> filterDeletedRegions(final Set<TableName> existingTables,
```
anew: Better name this filterDeletedTables(), or filterDeletedTableRegions()
- it does not filter out deleted regions
- it does filter out all regions of deleted tables

tephra-hbase-compat-1.1-base/src/test/java/org/apache/tephra/hbase/txprune/InvalidListPruneTest.java
```diff
@@ -384,7 +384,67 @@ public void testPruneEmptyTable() throws Exception {
...
+      // fetch prune upper bound, there should be a prune upper bound even though txTempTable does not exist anymore,
+      // and txDataTable2 has not been compacted/flushed yet
+      long now2 = System.currentTimeMillis();
+      long inactiveTxTimeNow2 = (now1 - 150) * TxConstants.MAX_TX_PER_MS;
+      long pruneUpperBound2 = transactionPruningPlugin.fetchPruneUpperBound(now2, inactiveTxTimeNow2);
+      Assert.assertEquals(expectedPruneUpperBound1, pruneUpperBound2);
+      transactionPruningPlugin.pruneComplete(now2, expectedPruneUpperBound1);
+    } finally {
+      transactionPruningPlugin.destroy();
+      if (txDataTable2 != null) {
```
anew: you should also do this for txTempTable. The test case deletes it, but if it fails before that, the table never gets deleted.

tephra-hbase-compat-1.1-base/src/test/resources/logback-test.xml
```diff
@@ -0,0 +1,39 @@
...
+
+<configuration>
+  <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
+    <encoder>
+      <pattern>%d{ISO8601} - %-5p [%t:%C{1}@%L] - %m%n</pattern>
+    </encoder>
+  </appender>
+
+  <logger name="org.apache.hadoop" level="WARN" />
+  <!-- BlockStateChange is used by org.apache.hadoop.hdfs.server.blockmanagement.BlockManager -->
```
anew: seems like this would be useful for all test cases... it removes a lot of distracting log messages.

tephra-hbase-compat-1.1-base/src/test/java/org/apache/tephra/hbase/txprune/InvalidListPruneTest.java
poornachandra: 👍

tephra-hbase-compat-1.1-base/src/test/resources/logback-test.xml
poornachandra: I'll copy over this file to the other compat modules

PR #58, opened 2017-09-13
TEPHRA-231 Fix LICENSE and NOTICE files for release distribution

gokulavasan:
JIRA : https://issues.apache.org/jira/browse/TEPHRA-231


poornachandra:
I did a build and was able to see the new license and notice files in the distribution jar.

Can we rename `LICENSE_RELEASE` to `LICENSE_DISTRIBUTION` and `NOTICE_RELEASE` to `NOTICE_DISTRIBUTION`, since they only apply to the distribution tarball? Also is it okay to move them into `tephra-distribution/` directory for the same reason? In that case we can just call them `LICENSE` and `NOTICE`

gokulavasan:
@poornachandra Thanks for the review. Addressed your comment (moved LICENSE and NOTICE files under tephra-distribution). Please take a look again.

PR #61, opened 2017-09-16
TEPHRA-263 Enforce TTL, regardless of any in-progress transactions.

anwar6953:
https://issues.apache.org/jira/browse/TEPHRA-263
Instead of filtering data that is older than `tx.getVisibilityUpperBound - TTL`, we now filter data that is `tx.getTransactionId - TTL`.

This PR also makes two additional changes:
1. A large TTL (50 years, for instance) previously caused an exception since HBase start timestamps can not be negative.
2. TTL is now more accurately applied for TransactionVisibilityFilter, where `readNonTxnData` is true.

I'll apply the diff to the other hbase compat modules once these changes pass review.

### DIFF
--- a/tephra-core/src/main/java/org/apache/tephra/util/TxUtils.java
+++ b/tephra-core/src/main/java/org/apache/tephra/util/TxUtils.java
@@ -19,6 +19,8 @@
 package org.apache.tephra.util;
 
 import com.google.common.primitives.Longs;
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
 import org.apache.tephra.Transaction;
 import org.apache.tephra.TransactionManager;
 import org.apache.tephra.TransactionType;
@@ -61,9 +63,7 @@ public class TxUtils {
    * @return The oldest timestamp that will be visible for the given transaction and TTL configuration
    */
   public static long getOldestVisibleTimestamp(Map<byte[], Long> ttlByFamily, Transaction tx) {
-    long maxTTL = getMaxTTL(ttlByFamily);
-    // we know that data will not be cleaned up while this tx is running up to this point as janitor uses it
-    return maxTTL < Long.MAX_VALUE ? tx.getVisibilityUpperBound() - maxTTL * TxConstants.MAX_TX_PER_MS : 0;
+    return getOldestVisibleTimestamp(ttlByFamily, tx, false);
   }
 
   /**
@@ -75,12 +75,28 @@ public class TxUtils {
    * @return The oldest timestamp that will be visible for the given transaction and TTL configuration
    */
   public static long getOldestVisibleTimestamp(Map<byte[], Long> ttlByFamily, Transaction tx, boolean readNonTxnData) {
-    if (readNonTxnData) {
-      long maxTTL = getMaxTTL(ttlByFamily);
-      return maxTTL < Long.MAX_VALUE ? System.currentTimeMillis() - maxTTL : 0;
+    long maxTTL = getMaxTTL(ttlByFamily);
+    if (maxTTL == Long.MAX_VALUE) {
+      return 0;
     }
+    return getOldestVisibleTimestamp(maxTTL, tx, readNonTxnData);
+  }
 
-    return getOldestVisibleTimestamp(ttlByFamily, tx);
+  /**
+   * Returns the oldest visible timestamp for the given transaction, based on the TTL configured.  If the TTL is
+   * negative or zero, the oldest visible timestamp will be {@code 0}.
+   * @param ttl TTL value (in milliseconds)
+   * @param tx The current transaction
+   * @param readNonTxnData indicates that the timestamp returned should allow reading non-transactional data
+   * @return The oldest timestamp that will be visible for the given transaction and TTL configuration
+   */
+  public static long getOldestVisibleTimestamp(long ttl, Transaction tx, boolean readNonTxnData) {
+    if (ttl <= 0) {
+      return 0;
+    }
+    long ttlFactor = readNonTxnData ? 1 : TxConstants.MAX_TX_PER_MS;
+    // if the computed ttl is negative, return 0 because timestamps can not be negative
+    return Math.max(0, tx.getTransactionId() - ttl * ttlFactor);
   }
 
   /**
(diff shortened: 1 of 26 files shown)


tephra-core/src/main/java/org/apache/tephra/util/TxUtils.java
```diff
@@ -75,12 +75,9 @@ public static long getOldestVisibleTimestamp(Map<byte[], Long> ttlByFamily, Tran
...
   public static long getOldestVisibleTimestamp(Map<byte[], Long> ttlByFamily, Transaction tx, boolean readNonTxnData) {
-    if (readNonTxnData) {
-      long maxTTL = getMaxTTL(ttlByFamily);
-      return maxTTL < Long.MAX_VALUE ? System.currentTimeMillis() - maxTTL : 0;
-    }
-
-    return getOldestVisibleTimestamp(ttlByFamily, tx);
+    long ttlFactor = readNonTxnData ? 1 : TxConstants.MAX_TX_PER_MS;
+    long maxTTL = getMaxTTL(ttlByFamily);
+    return maxTTL < Long.MAX_VALUE ? tx.getTransactionId() - maxTTL * ttlFactor : 0;
```
anew: this can be negative if a large TTL is used.

tephra-hbase-compat-1.0/src/test/java/org/apache/tephra/hbase/coprocessor/TransactionVisibilityFilterTest.java
```diff
@@ -290,44 +290,52 @@ private void runFilteringTest(TxFilterFactory txFilterFactory,
...
   public void testTTLFiltering() throws Exception {
     Map<byte[], Long> ttls = Maps.newTreeMap(Bytes.BYTES_COMPARATOR);
-    ttls.put(FAM, 10L);
-    ttls.put(FAM2, 30L);
+    ttls.put(FAM, 100L);
+    ttls.put(FAM2, 300L);
     ttls.put(FAM3, 0L);
 
+    // start a transaction to populate the in-progress list. It should not affect TTL calculations
+    txManager.startShort();
```
anew: abort this tx at the end of the test?

tephra-hbase-compat-1.0/src/test/java/org/apache/tephra/hbase/coprocessor/TransactionVisibilityFilterTest.java
```diff
@@ -290,44 +290,52 @@ private void runFilteringTest(TxFilterFactory txFilterFactory,
...
+
+    // commit a transaction, just to move the readPointer. Otherwise, the TransactionVisibilityFilter will filter
+    // based upon the readPointer being smaller than the tested values, whereas we want to test the TTL filtering
+    Transaction superShortTx = txManager.startShort();
```
anew: actually, you don't really need the txManager for this test. You can simply construct a transaction that has the desired properties.

tephra-core/src/main/java/org/apache/tephra/util/TxUtils.java
```diff
@@ -75,12 +75,9 @@ public static long getOldestVisibleTimestamp(Map<byte[], Long> ttlByFamily, Tran
...
    */
   public static long getOldestVisibleTimestamp(Map<byte[], Long> ttlByFamily, Transaction tx, boolean readNonTxnData) {
-    if (readNonTxnData) {
-      long maxTTL = getMaxTTL(ttlByFamily);
-      return maxTTL < Long.MAX_VALUE ? System.currentTimeMillis() - maxTTL : 0;
-    }
-
-    return getOldestVisibleTimestamp(ttlByFamily, tx);
+    long ttlFactor = readNonTxnData ? 1 : TxConstants.MAX_TX_PER_MS;
+    long maxTTL = getMaxTTL(ttlByFamily);
```
chtyim: the `maxTTL` should be capped by `Math.min(maxTTL, Long.MAX_VALUE / ttlFactor)`, right?

tephra-hbase-compat-1.0/src/test/java/org/apache/tephra/hbase/coprocessor/TransactionVisibilityFilterTest.java
anwar6953: Many other test cases don't complete the tx lifecycle by the end of their test cases.
I think this is fine because a new TransactionManager is used for each test case.

tephra-core/src/main/java/org/apache/tephra/util/TxUtils.java
anwar6953: I think it should instead get capped by a lower value. In any case, I pushed a change that modifies this code. Please take another look.

tephra-hbase-compat-1.0/src/main/java/org/apache/tephra/hbase/coprocessor/TransactionVisibilityFilter.java
```diff
@@ -89,13 +89,13 @@ public TransactionVisibilityFilter(Transaction tx, Map<byte[], Long> ttlByFamily
...
   public TransactionVisibilityFilter(Transaction tx, Map<byte[], Long> ttlByFamily, boolean allowEmptyValues,
-                              ScanType scanType, @Nullable Filter cellFilter) {
+                                     ScanType scanType, @Nullable Filter cellFilter) {
     this.tx = tx;
     this.oldestTsByFamily = Maps.newTreeMap();
     for (Map.Entry<byte[], Long> ttlEntry : ttlByFamily.entrySet()) {
       long familyTTL = ttlEntry.getValue();
       oldestTsByFamily.put(new ImmutableBytesWritable(ttlEntry.getKey()),
-                           familyTTL <= 0 ? 0 : tx.getVisibilityUpperBound() - familyTTL * TxConstants.MAX_TX_PER_MS);
+                           familyTTL <= 0 ? 0 : tx.getTransactionId() - familyTTL * TxConstants.MAX_TX_PER_MS);
```
poornachandra: We can have -ve timestamps here too

anew:
LGTM from my side, please wait for @poornachandra 's opinion, too, for the last commit.

anwar6953:
Got LGTM on changes for 1 compat module, so squashed commits.

anew:
Changes look the same to me in all modules. LGTM

anwar6953:
Squshed commmits.

PR #63, opened 2017-09-21
TEPHRA-265 Fix NOTICE_BINARY for Guice and Guice Assisted Inject libraries

gokulavasan:
JIRA : https://issues.apache.org/jira/browse/TEPHRA-265


gokulavasan:
@johnament Please review this PR that fixes the NOTICE file for Apache Tephra when you get a chance. Thank you!

tephra-distribution/NOTICE_BINARY
```diff
@@ -6,10 +6,19 @@ The Apache Software Foundation (http://www.apache.org/).
...
+The Apache Software Foundation (http://www.apache.org/).
+
+Apache Tephra includes Guice Assisted Inject v3.0.
+Google Guice - Extensions - AssistedInject
+Copyright 2006-2011 Google, Inc.
+
+This product includes software developed at
+The Apache Software Foundation (http://www.apache.org/).
 
 Apache Tephra includes Guava v13.0.1. Tephra elects to include this software
```
johnament: most of these lines are still not relevant.

gokulavasan: @johnament Should I remove these lines? I couldn't find a NOTICE in Guava, Gson, javax.inject JARs.

poornachandra:
@gokulavasan Do you have time to take this PR to completion?

tephra-distribution/NOTICE_BINARY
johnament: Just because a dependency is apache licensed doesn't mean it will have a NOTICE.  NOTICE is optional, so you shouldn't include one if none is present.

gokulavasan:
@johnament Thanks for the review John. Please take another look when you get a chance.

poornachandra:
@johnament We are planning for the next release of Tephra soon. Can you take a look at the changes, so that this PR can be part of the next release?

PR #67, opened 2017-12-18
TEPHRA-272 Add HBase 2.0 compatibility module

ankitsinghal:


poornachandra:
@ankitsinghal I'm taking a look at the PR. Meanwhile, the Travis build has failed. Looks like there is some issue finding HBase 2.0 dependencies. Can you take a look at it?

Also, I'm on vacation this month. So there may be a delay in my replies.

tephra-hbase-compat-2.0/src/main/java/org/apache/tephra/hbase/TransactionAwareHTable.java
```diff
@@ -0,0 +1,729 @@
...
+        op.setAttribute(TxConstants.TX_OPERATION_ATTRIBUTE_KEY, txCodec.encode(tx));
+    }
+
+    protected void makeRollbackOperation(Delete delete) {
+        delete.setAttribute(TxConstants.TX_ROLLBACK_ATTRIBUTE_KEY, new byte[0]);
+    }
+
+    @Override
+    public TableDescriptor getDescriptor() throws IOException {
+        // TODO Auto-generated method stub
```
poornachandra: Missing implementation?

tephra-hbase-compat-2.0/src/main/java/org/apache/tephra/hbase/TransactionAwareHTable.java
```diff
@@ -0,0 +1,729 @@
...
+        return null;
+    }
+
+    @Override
+    public boolean[] exists(List<Get> gets) throws IOException {
```
poornachandra: It would be good to move all the public methods before the private helpers for readability.

tephra-hbase-compat-2.0/src/main/java/org/apache/tephra/hbase/TransactionAwareHTable.java
```diff
@@ -0,0 +1,729 @@
...
+        }
+        List<Get> transactionalizedGets = new ArrayList<>(gets.size());
+        for (Get get : gets) {
+            transactionalizedGets.add(transactionalizeAction(get));
+        }
+        return hTable.exists(transactionalizedGets);
+    }
+
+    @Override
+    public boolean checkAndPut(byte[] row, byte[] family, byte[] qualifier, CompareOperator op,
```
poornachandra: Would be good to move the `checkAnd...` methods here alongside the `checkAnd...` earlier in the file.

tephra-hbase-compat-2.0/src/main/java/org/apache/tephra/hbase/coprocessor/TransactionProcessor.java
```diff
@@ -0,0 +1,583 @@
...
+ * {@code
+ * Transaction t = ...;
+ * Get get = new Get(...);
+ * TransactionCodec codec = new TransactionCodec();
+ * codec.addToOperation(get, t);
+ * }
+ * </p>
+ */
+public class TransactionProcessor implements RegionObserver, RegionCoprocessor {
+    private static final Log LOG = LogFactory.getLog(TransactionProcessor.class);
```
poornachandra: This file also needs to use 2 spaces for indentation

tephra-hbase-compat-2.0/src/test/java/org/apache/tephra/hbase/AbstractHBaseTableTest.java
```diff
@@ -0,0 +1,106 @@
...
+      desc.setValue(TxConstants.READ_NON_TX_DATA, "true");
+    }
+    // Divide individually to prevent any overflow
+    int priority = Coprocessor.PRIORITY_USER;
+    // order in list is the same order that coprocessors will be invoked
+    for (String coprocessor : coprocessors) {
+      desc.addCoprocessor(coprocessor, null, ++priority, null);
+    }
+    hBaseAdmin.createTable(desc);
+    //testUtil.waitTableAvailable(tableName, 5000);
```
poornachandra: Is the wait no longer required here?

tephra-hbase-compat-2.0/src/test/java/org/apache/tephra/hbase/TransactionAwareHTableTest.java
```diff
@@ -0,0 +1,1907 @@
...
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertTrue;
+import static org.junit.Assert.fail;
+
+
+/**
+ * Tests for TransactionAwareHTables.
+ */
+public class TransactionAwareHTableTest extends AbstractHBaseTableTest {
+    private static final Logger LOG = LoggerFactory.getLogger(TransactionAwareHTableTest.class);
```
poornachandra: This file also needs to use 2 space for indentation

tephra-hbase-compat-2.0/src/test/java/org/apache/tephra/hbase/coprocessor/TransactionProcessorTest.java
```diff
@@ -0,0 +1,677 @@
...
+    HDFSTransactionStateStorage tmpStorage =
+      new HDFSTransactionStateStorage(conf, new SnapshotCodecProvider(conf), new TxMetricsCollector());
+    tmpStorage.startAndWait();
+    tmpStorage.writeSnapshot(txSnapshot);
+    tmpStorage.stopAndWait();
+    long globalMemStoreLimit = (long) (ManagementFactory.getMemoryMXBean().getHeapMemoryUsage()
+            .getMax() * MemorySizeUtil.getGlobalMemStoreHeapPercent(conf, false));
+    chunkCreator = ChunkCreator.initialize(MemStoreLABImpl.CHUNK_SIZE_DEFAULT, false,
+            globalMemStoreLimit, 0.2f, MemStoreLAB.POOL_INITIAL_SIZE_DEFAULT, null);
+    assertTrue(chunkCreator != null);
```
poornachandra: I don't see `chunkCreator` being used anywhere else in the file after the initialization here. Is this required?

tephra-hbase-compat-2.0/src/test/java/org/apache/tephra/hbase/coprocessor/TransactionProcessorTest.java
```diff
@@ -0,0 +1,677 @@
...
+
+      results.clear();
+      assertTrue(regionScanner.next(results));
+      assertKeyValueMatches(results, 7, new long[]{V[6], V[4]});
+
+      results.clear();
+      assertFalse(regionScanner.next(results));
+      assertKeyValueMatches(results, 8, new long[] {V[8], V[6], V[4]});
+    } finally {
+      //region.close();
```
poornachandra: Any reason for commenting out the `region.close()` here and in other places of this file?

tephra-hbase-compat-2.0/src/test/java/org/apache/tephra/hbase/coprocessor/TransactionProcessorTest.java
```diff
@@ -0,0 +1,677 @@
...
+          // deletes are performed as puts with empty values
+          Put deletePut = new Put(row);
+          deletePut.addColumn(familyBytes, Bytes.toBytes(i), deleteTs, new byte[0]);
+          region.put(deletePut);
+        }
+      }
+      
+      // read all back
+      scan = new Scan(row);
+      scan.readVersions(1);
```
poornachandra: What happens if `scan.readVersions(1)` is not set here and in line 358?

tephra-hbase-compat-2.0/src/test/java/org/apache/tephra/hbase/coprocessor/TransactionProcessorTest.java
```diff
@@ -0,0 +1,677 @@
...
+  @SuppressWarnings("StatementWithEmptyBody")
+  private void scanAndAssert(HRegion region, List<Cell> expected, Scan scan) throws Exception {
+    try (RegionScanner regionScanner = region.getScanner(scan)) {
+      List<Cell> results = Lists.newArrayList();
+      while (regionScanner.next(results)) { }
+      assertEquals(expected, results);
+    }
+  }
+
+    private HRegion updateTtl(HRegion region, byte[] family, long ttl) throws Exception {
```
poornachandra: This method needs to use 2 spaces for indentation.

tephra-hbase-compat-2.0/src/test/java/org/apache/tephra/hbase/coprocessor/TransactionProcessorTest.java
```diff
@@ -0,0 +1,677 @@
...
+                TableDescriptorBuilder.newBuilder(region.getTableDescriptor());
+        ColumnFamilyDescriptorBuilder cfd =
+                ColumnFamilyDescriptorBuilder
+                        .newBuilder(tableBuilder.build().getColumnFamily(family));
+        if (ttl > 0) {
+            cfd.setValue(Bytes.toBytes(TxConstants.PROPERTY_TTL),
+                Bytes.toBytes(String.valueOf(ttl)));
+        }
+        cfd.setMaxVersions(10);
+        tableBuilder.removeColumnFamily(family);
```
poornachandra: Just curious as to why we need to remove the column family and then add it back again?

ankitsinghal: Actually, when we alter the existing column family and then we need to remove and add it otherwise builder will say column family already exists

tephra-hbase-compat-2.0/src/test/java/org/apache/tephra/hbase/coprocessor/TransactionProcessorTest.java
ankitsinghal: We need to initialize ChunkCreater as it is used internally by HRegion while creating chunk.

ankitsinghal:
Thanks a lot @poornachandra for the review, I have taken care of all the current review comments in the last commit. Let me know if any changes required with your further review.

tephra-hbase-compat-2.0/src/main/java/org/apache/tephra/hbase/coprocessor/FilteredInternalScanner.java
```diff
@@ -0,0 +1,80 @@
...
+  public boolean next(List<Cell> result, ScannerContext scannerContext) throws IOException {
+    List<Cell> outResult = new ArrayList<Cell>();
+    while (true) {
+      boolean next = delegate.next(outResult, scannerContext);
+      for (Cell cell : outResult) {
+        ReturnCode code = filter.filterKeyValue(cell);
+        switch (code) {
+        // included, so we are done
+        case INCLUDE:
+        case INCLUDE_AND_NEXT_COL:
```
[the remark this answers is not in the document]
ankitsinghal: Yes, so on next cell, we expect the filter to return some NEXT_COL/SKIP (if it returned INCLUDE_AND_NEXT_COL for the previous cell) so that we will skip them.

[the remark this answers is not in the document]
poornachandra: 👍  Just confirmed that transaction filter does this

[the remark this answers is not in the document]
poornachandra: The only downside I see with this approach is that we'll iterate through all the cells for a column, even though we could have skipped some cells when the filter returns `INCLUDE_AND_NEXT_COL`. 

For example if there are 100 versions of a value in a column, and the filter figures out at the 10th version that the rest of the versions have reached the TTL, then in the earlier code we would not  iterate through the rest of the 90 cells (or at least that's what I think was happening in the underlying scanner). Is there anyway we can achieve the same result with the new APIs?

poornachandra:
@ankitsinghal Just one more question (https://github.com/apache/incubator-tephra/pull/67#discussion_r169503415). 

Also the Java 7 build is failing for HBase 2.0 compat module. Looks like HBase 2.0 needs Java 8. Can you update the pom for the 2.0 compat module to use Java 8? 

Note that the master branch has some changes on how the Travis tests are run (TEPHRA-282). You'll have to rebase, and then add an entry to travis.yml for the 2.0 compat module.

tephra-hbase-compat-2.0/src/main/java/org/apache/tephra/hbase/coprocessor/FilteredInternalScanner.java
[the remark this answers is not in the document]
ankitsinghal: Yes @poornachandra , that's a downside, but with HBase-2.0 we don't have access to create StoreScanner with the custom scan in pre-compact/pre-flush hooks. 
I had a discussion with Anoop(From HBase) on https://issues.apache.org/jira/browse/TEPHRA-274 for reviewing the approach, probably currently there is no other way.

ankitsinghal:
bq.  Can you update the pom for the 2.0 compat module to use Java 8?
Do we want to make Java8 specific to HBase-2.0 compat module and keep other still running on Java7. Will it not affect the release process, as currently the release is built with single version JDK 1.7?

poornachandra:
@ankitsinghal Since HBase-2.0 needs Java 8, let's try changing the compiler for only HBase-2.0 compat module to use Java 8. We can still keep the other modules as Java 7. I haven't tried it out myself. But I think it should work.

Right now, we run Travis tests using both Java 7 and Java 8 for all modules. We can restrict HBase-2.0 compat module tests to run only for Java 8.

joshelser:
> We can still keep the other modules as Java 7. I haven't tried it out myself. But I think it should work.

@poornachandra is there a reason that Tephra is still on Java7? It's quite out-dated at this point..

poornachandra:
@joshelser There is nothing stopping us from moving Tephra to Java 8. Just that we'll have to build consensus on the Tephra dev list, and it will add to the delay of this PR. If making the POM changes to work with both Java 7 and Java 8 is complicated, then @ankitsinghal can you send out an email to the dev list for the Java 8 consensus?

ankitsinghal:
@poornachandra , I have already made the change to compile HBase-2.0 with Java 8. should I rebase and squash the commits for merge now?
It seems HBase 1.x releases are still with Java 7 so We can also support Java 7 for the older releases of HBase so that user can still use latest releases of Tephra without changing their environment. Though not sure why 1-2 tests of HBase 1.4 are failing on Java 7, we can fix this in separate JIRA.

ankitsinghal:
@poornachandra , have squashed all commit in a single one. can you please commit this.

poornachandra:
@ankitsinghal I spent some time on the tephra-hbase-compat-1.4 compat module tests failure. It looks like the 1.4 compat module tests were never run earlier. When you changed the travis.xml to build hbase-compat-1.4 module, the issue was exposed.

I have filed JIRA https://issues.apache.org/jira/browse/TEPHRA-285 for this issue. Since the failure was already present, I will go ahead and merge this PR. Thanks for the contribution!

PR #68, opened 2018-01-10
TEPHRA-278 Support HBase 1.4

apurtell:


anew:
@apurtell Is this a 1:1 copy of the hbase-1.3 compat module? Or is there an API change that requires code change in the tephra module? If 1.4 is API compatible with 1.3, we can simply add support for 1.4 by adding a rule to HBaseVersion.java and use the 1.3 compat module if the HBase version string  starts with "1.4"

apurtell:
There are 4 new methods in the abstract interface Table:

    int getReadRpcTimeout();
    void setReadRpcTimeout(int readRpcTimeout);
    int getWriteRpcTimeout();
    void setWriteRpcTimeout(int writeRpcTimeout);

Tephra won't build without changes that implement these.

apurtell:
I didn't know about HBaseVersion.java. Added a new conditional to that just now.

anew:
Changes look good to me - but both Travis runs failed with the error:
```
testFamilyDeleteWithCompaction(org.apache.tephra.hbase.TransactionAwareHTableTest)  Time elapsed: 69.08 sec  <<< FAILURE!
java.lang.AssertionError: Compaction should have removed the row
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.assertTrue(Assert.java:41)
	at org.apache.tephra.hbase.TransactionAwareHTableTest.testFamilyDeleteWithCompaction(TransactionAwareHTableTest.java:491)
```
Since this PR does not change any of the coprocessor logic, I am not sure what would cause this. Do you know whether anything changed in HBase 1.4 that would change the behavior of compaction?

apurtell:
Yes this could be an issue with 1.4. How about merging this PR so we can get Phoenix builds with 1.4 working for some definition of that and open another issue for the test failure? I can look at that next?

anew:
Running this test locally, I can see the cause of the failure in the logs:
```
2018-01-11 15:10:52,501 - INFO  [TransactionStateCache STARTING:o.a.t.c.TransactionStateCache@113] - Failed to initialize TransactionStateCache due to: 
com.google.inject.ProvisionException: Guice provision errors:

1) Unknown HBase version: 1.4.0

1 error
	at org.apache.tephra.util.HBaseVersionSpecificFactory.get(HBaseVersionSpecificFactory.java:60) ~[classes/:na]
	at org.apache.tephra.util.ConfigurationFactory.<init>(ConfigurationFactory.java:66) ~[classes/:na]
	at org.apache.tephra.coprocessor.TransactionStateCache.getSnapshotConfiguration(TransactionStateCache.java:118) [classes/:na]
	at org.apache.tephra.coprocessor.TransactionStateCache.tryInit(TransactionStateCache.java:99) [classes/:na]
	at org.apache.tephra.coprocessor.TransactionStateCache.refreshState(TransactionStateCache.java:158) [classes/:na]
	at org.apache.tephra.coprocessor.TransactionStateCache.startUp(TransactionStateCache.java:75) [classes/:na]
	at com.google.common.util.concurrent.AbstractIdleService$1$1.run(AbstractIdleService.java:43) [guava-13.0.1.jar:na]
	at java.lang.Thread.run(Thread.java:745) [na:1.7.0_80]
```
So the fix would be to 
- add a case for 1.4 in HBaseVersionSpecificFactory.get(). 
- that will require also adding a getHBase14Classname() and implementing it in ConfigurationFactory. 
- rename the HBase13ConfigurationProvider in the 1.4 compat module to HBase14ConfigurationProvider (along with its test case). 

I tried this in my local build and the test then passes.

apurtell:
Thanks so much for the hints. Still away from a computer. When I’m back I will fix this. Soon

apurtell:
Almost there, need to fix this

    Failed tests: 
      HBase14ConfigurationProviderTest>AbstractConfigurationProviderTest.testVersionFactory:37 HBASE_14 was not part of the expected versions : [HBASE_13]

apurtell:
The CI builds are terminating due to this

    The job exceeded the maximum time limit for jobs, and has been terminated.

The build completes, including all tests, locally

    [INFO] BUILD SUCCESS
    [INFO] ------------------------------------------------------------------------
    [INFO] Total time: 47:56 min
    [INFO] Finished at: 2018-01-12T10:46:22-08:00
    [INFO] Final Memory: 248M/4334M
    [INFO] ------------------------------------------------------------------------

Took 48 minutes. Maybe the CI config needs adjusting?

PR #69, opened 2018-01-12
(TEPHRA-279) Make TransactionContext resilient to exceptions from getTransactionAwareName()

anew:


anew:
rebased on current master - waiting for one more build before committing this.

anew:
One test failure, that was present before this. Will commit this now.

PR #70, opened 2018-01-16
[TEPHRA-282] Split Travis build per compat module

anew:


anew:
Travis was down yesterday - initiating another build now by force-pushing a slightly different commit

anew:
So this worked, but triggered 20 builds: https://travis-ci.org/apache/incubator-tephra/builds/330008721. There is a lot of redundancy in these tests, because each of them builds and tests tephra-core. I am gonna try to consolidate this a little by combining:

- 0.96 and 0.98 
- 1.0 and 1.0-cdh 
- 1.1, 1.2, cdh-5.7 and cdh-5.8 (they all share compat-hbase-1.1-base)
- 1.3 and 1.4

That reduces it to 2x4 builds and hopefully each one will pass way below the 50 minute mark.

anew:
@apurtell based on our discussion - what do you think of this approach?

apurtell:
looks good to me!

anew:
Thanks @apurtell !

PR #71, opened 2018-03-23
(TEPHRA-285) Exclude failing test from Travis build 

anew:
…until fix is available in HBase


poornachandra:
@anew It would be good to commit this against JIRA TEPHRA-285, looks like you intended to use that but ended up using TEPHRA-275

anew:
@poornachandra you are right. Changed the commit message.

anew:
I had to try a few options:

- -Dtest=*,!TransactionProcessorTest does not seem to work in travis (even though it worked fine on my Mac). It then skips all tests and fails because of that
- so I added an extra exclude rule in the pom file
- I could not figure out how to modify a single build in the matrix
- excluded the regular jdk7/hbase-1.4 build, and included it with an extra test exclude

All builds now pass and only the jdk7/hbase-1.4 build skips TransactionProcessorTest

PR #74, opened 2018-04-26
TEPHRA-266 Identify log messages when multiple instances of Tephra run on a single HBase cluster

poornachandra:
JIRA - https://issues.apache.org/jira/browse/TEPHRA-266

### DIFF
--- a/tephra-core/src/main/java/org/apache/tephra/coprocessor/TransactionStateCache.java
+++ b/tephra-core/src/main/java/org/apache/tephra/coprocessor/TransactionStateCache.java
@@ -55,6 +55,7 @@ public class TransactionStateCache extends AbstractIdleService implements Config
   // snapshot refresh frequency in milliseconds
   private long snapshotRefreshFrequency;
   private boolean initialized;
+  private String logPrefix = "";
 
   public TransactionStateCache() {
   }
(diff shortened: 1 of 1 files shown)


tephra-core/src/main/java/org/apache/tephra/coprocessor/TransactionStateCache.java
```diff
@@ -184,4 +185,14 @@ private void refreshState() throws IOException {
...
     return latestState;
   }
+
+  protected void setId(@Nullable String id) {
+    if (id != null) {
+      this.logPrefix = "[" + id + "] ";
+    }
+  }
+
+  private String prefixLog(String message) {
```
anew: not sure this is a very good idea. It means you are performing the string operations even when it is not being logged (for example, for debug messages). Better to add the logPrefix as an argument to the log message, such as:
```
LOG.debug("[{}] Latest transaction snapshot: {}", logPrefix, latestState.toString()));
```

poornachandra: HBase co-processor uses Apache commons-logging. The `{}` notation is not available in the commons-logging APIs.

I couldn't figure out a way to not create a string unless absolutely necessary. The only debug log in that file has `LOG.isDebugEnabled()` guard. However, I know this is not an ideal solution.

anew: I was not aware of that. I think this is not logged very frequently, so it's ok.

PR #75, opened 2018-05-04
TEPHRA-287 ActionChange.getChangeKey() not implemented correctly

JamesRTaylor:
Based on a new config parameter (defaulting to true), send the old change set row key and the new change set row key. This allows a mix of old and new clients to still detect conflicts across them. Once it's know that only new clients are being used, the config parameter can be switched to false in which case only the new change set row key (which correctly embeds separator bytes to prevent false positives) will be sent.

### DIFF
--- a/tephra-core/src/main/java/org/apache/tephra/AbstractTransactionAwareTable.java
+++ b/tephra-core/src/main/java/org/apache/tephra/AbstractTransactionAwareTable.java
@@ -24,6 +24,8 @@ import com.google.common.collect.Sets;
 import com.google.common.primitives.Bytes;
 import com.google.common.primitives.UnsignedBytes;
 
+import org.apache.hadoop.io.WritableUtils;
+
 import java.io.IOException;
 import java.util.Arrays;
 import java.util.Collection;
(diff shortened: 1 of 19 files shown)


tephra-core/src/main/java/org/apache/tephra/AbstractTransactionAwareTable.java
```diff
@@ -88,21 +93,105 @@ public void updateTx(Transaction tx) {
...
+    }
+    return result;
+  }
+
+  /**
+   * The unique bytes identifying what is changing. We use the
+   * following structure:
+   * ROW conflict level: <table_name><0 byte separator><row key>
+   * since we know that table_name cannot contain a zero byte.
+   * COLUMN conflict level: <table_name><length of family as vint><family>
```
poornachandra: Based on the implementation below this should be `COLUMN conflict level: <table_name>0<length of family as vint><family>...`

tephra-core/src/main/java/org/apache/tephra/AbstractTransactionAwareTable.java
```diff
@@ -88,21 +93,105 @@ public void updateTx(Transaction tx) {
...
+  private byte[] getChangeKeyWithoutSeparators(byte[] row, byte[] family, byte[] qualifier) {
     byte[] key;
     byte[] tableKey = getTableKey();
     switch (conflictLevel) {
     case ROW:
       key = Bytes.concat(tableKey, row);
       break;
     case COLUMN:
-      key = Bytes.concat(tableKey, row, family, qualifier);
+      key = Bytes.concat(tableKey, family, qualifier, row);
```
poornachandra: Position of `row` in the change key has changed. Looks like this was unintentional?

PR #101, opened 2020-07-27
TEPHRA-307 Support Hbase 2.3

stoty:


stoty:
This is a bit more involved than the 2.1+2.2 work was, as there are API changes (extra method in Table interface), so this is an actual new module.

PR #103, opened 2020-08-05
Tephra 309

stoty:


tephra-hbase-compat-1.4-base/pom.xml
```diff
@@ -0,0 +1,221 @@
...
+                <goal>add-test-source</goal>
+              </goals>
+              <configuration>
+                <sources>
+                  <source>${project.parent.basedir}/src/test/java</source>
+                </sources>
+              </configuration>
+            </execution>
+          </executions>
+        </plugin>
```
joshelser, line 210: Shouldn't be necessary -- these are included by default.

stoty: This section copies the common java source files from -common to the version specific sub-sub-projects that have no source files at all. See the <sources> element. The same approach is used in all the -common based compatibility modules that share the java code.

stoty:
> `tephra-examples/tephra-examples-post-1.3/hbase-1.6/pom.xml~` shouldn't be staged?

Of course. Thanks for noticing it.

stoty:
removed backup files, rebased, and re-added shaded artifacts.

tephra-hbase-compat-1.4-base/pom.xml
joshelser: OH! I missed that this is `project.PARENT.basedir`. Thanks.

PR #108, opened 2020-11-12
TEPHRA-313 Remove support for ancient HBase releases

stoty:
remove support for HBase veresions older than 1.1


anew (approved):
Changes look good. 

After removing all those old versions, a lot of stuff in the compatibility modules is probably redundant. But that's a more involved refactoring that can wait until 1.1 and 1.2 are removed, too.

PR #109, opened 2020-11-12
TEPHRA-314 Provide a sensible tephra server assembly

stoty:


anew (approved):
nothing looks wrong about this.

stoty:
Thank you. 
I have renamed the ticket and the commit message, as this ended up smaller in scope than I originally intended,

PR #112, opened 2021-04-16
TEPHRA-317 Add support for HBase 2.4

richardantal:


stoty:
1. There seems to be a pom error that causes the test to fail.

2. Did you have to make changes in the code for 2.4 compared to 2.3 ? If not, then it'd be better to turn the 2.3 module into 2.3-base, and use that as a base for the actual 2.3 and 2.4 modules to avoid code duplication. (see the 2.0-base module for an example)

richardantal:
> 1. There seems to be a pom error that causes the test to fail.
> 2. Did you have to make changes in the code for 2.4 compared to 2.3 ? If not, then it'd be better to turn the 2.3 module into 2.3-base, and use that as a base for the actual 2.3 and 2.4 modules to avoid code duplication. (see the 2.0-base module for an example)

1. Thanks, tephra-examples/tephra-examples-post-1.3/hbase-2.4/pom.xml was missing, locally it was fine but I forgot to add it to git.

2. I made 2 small code changes. One of them was in TransactionAwareHTable, there is a mutateRow function. It was void but now it returns Result type.
The second change is in the TransactionProcessorTest, the ChunkCreator function has an additional indexChunkSizePercent parameter.

stoty:
Thanks.
Let's wait for the CI run.

stoty (approved):
+1 LGTM

